PR #37, opened 2024-03-30
Add import from splitwise option

### LINKED ISSUE #29: Import data from splitwise.
Currently export option is only available for splitwise pro users. Need to figure out way to make it wotk for everyone!

KMKoushik:
Fixes #29


dfego:
Woohoo! 🎉 Is this in the live version on the Web and I haven't found the button or is it not yet deployed?

PR #62, opened 2024-05-15
[FEATURE] Self Hosting using docker, docker compose, nginx, certbot and backups using pg_dump and mutt

vishwa5854:
Although this is already hosted on vercel right now, enthusiasts like me, generally like to host their own version of it. Hence the PR.
1. Added support for custom db migration which will build a custom postgres image
2. Added support for spinning up containers in a self hosted env with one command
3. Updated instructions, env.example
4. Added support for periodic backups of db and send them as email


prisma/seed.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
KMKoushik: Just a question, why can't we use the prisma's migration command?

KMKoushik (changes requested):
TYSM, This is really awesome. I have a one small question

prisma/seed.sh
vishwa5854: Because at this point of time there is no postgres instance running, we are in the build phase of the postgres instance itself and Prisma can only work if there is an active postgres instance that's listening to requests, where as this script merges all the SQL migrations into one SQL file and puts it in the init folder of docker container which is automatically executed as soon as the postgres container is started for the first time.

KMKoushik: In that case we can do the migration in app dockerfile right? I'm just worried we are running 2 diff code for migration. And prisma migration handles lot's of cases. For instance let's I want to update the server which has migrations. This docker script will fail

vishwa5854: Yeah that makes sense, if the db already has old migrations this script will fail. Thanks for pointing that out!

It wouldn't work by putting in the Docker file of the server, since the Dockerfile wouldn't know if the postgres container is running or not. 

I will see if I can do something in the compose using init scripts in the server container instead of postgres.

vishwa5854: @KMKoushik  What happens if I run `prisma:prod` multiple times? will it rerun all the migrations or will it only run any new migrations after the last run ? 
If it doesn't rerun the migrations again and again, then your suggestion will work since I have a `depends_on` attrib for the server, we can just put the prisma migration command in the start command instead of build command.

KMKoushik: Yep, it will run only the new migrations. It takes care of everythin

vishwa5854: Updated @KMKoushik

KMKoushik (approved):
This is awesome. TYSM

PR #70, opened 2024-06-27
change user groups icon to be a Plus sign

### LINKED ISSUE #68: [UX] Add group button icon misleading
The add group button icon is kind of misleading, since it has a user icon and + symbol, could confuse the user whether he is adding a user or a group
image
A simple + icon would suffice.

brenol:
fixes #68


KMKoushik (approved):
LGTM, thanks

Issue #96, opened 2024-09-08
Feature request: Unselect all when adding expense


jaclavij:
Hi, I recently started using split-pro, migrating from Tricount with a large group of 15 people.

It would be helpful to have an "Unselect all" button/icon when adding an expense in the SplitEqualSection, since we usually have expenses that only affect 2-5 people. Thus avoiding having to manually unselect 10 people, saving clicks.

I guess it also makes sense that it turns into "Select all" after unselecting.

PS: Thank you for this great app!

jaclavij:
I have started to work on this @KMKoushik

KMKoushik:
legend! thank you!

PR #101, opened 2024-09-10
Select/Unselect All in SplitEqualSection

jaclavij:
Adds #96 and fixes the vertical alignment of checks and separation between participants to be in line with the rest of the split types.


KMKoushik (approved):

KMKoushik:
TYSM

Issue #100, opened 2024-09-10
Feature Request: Statistics on group level


kruemelro:
It would be a great addition to have a statistics or summary page on group level. I have something in mind like

- total costs
- costs grouped by category
- costs grouped by day
- costs grouped by user

krokosik:
This looks like it will be partially covered by #197

krokosik:
Closing this as group balances were implemented in 1.4, while there is ongoing discussion #230 about statistics

PR #116, opened 2024-10-04
Show total expenses of a group

FriesischScott:
This PR adds a display of the total expenses in a group to the  group detail page.

My use case is this: I use a group to track the shared expenses for a holiday trip and am interested in the overall  expenses as a reference for future trips of the same kind.

Something similar was requested in #100.

I added the total costs as a second  `flex` column next to the amounts you lent or owe.
 
image

I don't have much experience with typescript but I'm confident that I got the types right and have verified that `pnpm build` runs successfully.


KMKoushik:
this is good, i'll be away this week. so i'll start merging them next week along with others

FriesischScott:
Great! I'll be happy to make any changes you want me to.

KMKoushik (changes requested):
the only problem with is, the total will keep on growing if you use same group, think it's better to have it in a separate section, so if users want it they can click and check them.

it can be a section here
Screenshot 2024-10-14 at 4 51 37 pm

FriesischScott:
> the only problem with is, the total will keep on growing if you use same group, think it's better to have it in a separate section, so if users want it they can click and check them.
> 
> it can be a section here Screenshot 2024-10-14 at 4 51 37 pm

If you think it's too much on the group detail page I'm happy to put it somewhere else.

The image you shared is of the `FriendBalance`. How would the total expenses of a group fit there?

Maybe I could link from the group detail page to a separate group statistics page?

KMKoushik:
> If you think it's too much on the group detail page I'm happy to put it somewhere else.
> 
> The image you shared is of the `FriendBalance`. How would the total expenses of a group fit there?
> 
> Maybe I could link from the group detail page to a separate group statistics page?

oh i meant as a button in the top section of the group details page, on clicking it can show a model!

FriesischScott:
> > If you think it's too much on the group detail page I'm happy to put it somewhere else.
> > The image you shared is of the `FriendBalance`. How would the total expenses of a group fit there?
> > Maybe I could link from the group detail page to a separate group statistics page?
> 
> oh i meant as a button in the top section of the group details page, on clicking it can show a model!

Sounds good. I'll do it.

FriesischScott:
I moved the total expenses to a new *Group statistics* modal. This way it could be extended with more statistics in the future.

You were right, The UI  ismuch cleaner this way.  I wasn't quite sure what icon to use so I opted for a simple bar chart to represent the statistics.

I also cleaned up a few unused variables and replaced the info icon with one from lucide. The old icon was quite thin compared to the lucide one.

KMKoushik (approved):
TYSM

PR #127, opened 2024-10-16
remove the need for auth provider env

KMKoushik:


KMKoushik:
@Striffly removed the need of `AUTH_PROVIDERS` here. jus letting you know

PR #89, opened 2024-08-30
implemented i18n

kruemelro:
I've implemented i18n translation framework and also added English and German as language. I'm no developer, so please have a look over it

PR #132, opened 2024-10-26
Add bank account transaction integration

alexanderwassbjer:
Implemented an integration that uses GoCardless api to fetch bank account transactions. Then when you want to add an expense you can either fill it manually or you can pick from your transactions and it will automatically be added as an expense. There is also a function to "multi add" multiple transactions at the same time.
If a transaction already has been added to a group you cant add it once again.
[GoCardless page](https://gocardless.com/bank-account-data/)

Other changes:
- Fixed dark/light mode.
- Fixed "total expenses" which didn't excluded deleted expenses.
- Added a clear field button to add expense page.
- Added the new feature to start page.
- Added information about the new feature to readme.

Screenshots:
image
image
image


alexanderwassbjer:
@KMKoushik :)

alexanderwassbjer:
I have also added i18n support for this feature in another PR that is based on the based on [this PR](https://github.com/oss-apps/split-pro/pull/89). If this gets merged I can open a PR with i18n support for this aswell.

.env.example
```diff
@@ -41,6 +41,12 @@ EMAIL_SERVER_PORT=
 EMAIL_SERVER_USER=
 EMAIL_SERVER_PASSWORD=
 
+# GoCardless options
```
KMKoushik, line 49: we can't use NEXT_PUBLIC cuz this won't work on docker builds check the notes section in this: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser

i recommend to have `GOCARDLESS_ENABLED`  and get it from the server

.env.example
```diff
@@ -41,6 +41,12 @@ EMAIL_SERVER_PORT=
...
+NEXT_PUBLIC_GOCARDLESS_ENABLED=
+NEXT_PUBLIC_GOCARDLESS_COUNTRY=
```
KMKoushik: so it will work on one specific country alone?

Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM node:20.11.1-alpine AS base
```
KMKoushik: we already have docker file under docker folder

docker-compose.yml
```diff
@@ -0,0 +1,65 @@
+name: split-pro-prod
```
KMKoushik: same we already have a file under docker folder

docker/Dockerfile
```diff
@@ -13,6 +13,11 @@ COPY package.json pnpm-lock.yaml ./
 
 RUN pnpm install
 
+ARG NEXT_PUBLIC_GOCARDLESS_ENABLED
```
KMKoushik: thing is i'm hosting prebuilt docker images and this should not be a part of docker image, rather should come run time!

src/pages/_app.tsx
```diff
@@ -64,7 +64,12 @@ const MyApp: AppType<{ session: Session | null }> = ({
         <meta property="og:image" content="https://splitpro.app/og_banner.png" />
       </Head>
       <SessionProvider session={session}>
-        <ThemeProvider attribute="class" defaultTheme="dark">
+        <ThemeProvider
```
KMKoushik: it is best to do this in separate PR and on option to swithch in account settings

KMKoushik (changes requested):
this looks good overall with some comments. please lemme know what you think!

tysm for this PR really appreciate it. this is really awesome!

think it's best to merge in v2, test thoroughly and release it with other features! (i'll create a new branch)

KMKoushik:
@alexanderwassbjer  changed the base to v2!

src/components/AddExpense/GoCardlessTransactions.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  if (!env.NEXT_PUBLIC_GOCARDLESS_ENABLED) {
+    return <></>;
+  }
+
+  const transactionsArray = returnTransactionsArray();
+
+  const onTransactionRowClick = (item: TransactionWithPendingStatus, multiple: boolean) => {
+    if (multiple) {
+      if (multipleArray?.some((cItem) => cItem.transactionId === item.transactionId)) {
```
frdrkolsson: Here

src/components/AddExpense/GoCardlessTransactions.tsx
```diff
@@ -0,0 +1,170 @@
...
+                  <Checkbox
+                    checked={multipleArray?.some(
+                      (cItem) => cItem.transactionId === item.transactionId,
+                    )}
+                    disabled={alreadyAdded(item.transactionId)}
+                    onCheckedChange={() => {
+                      onTransactionRowClick(item, true);
+                    }}
+                  />
+                  <button
```
frdrkolsson: Why not 
```suggestion
                  <Button
```

src/server/api/routers/nordigen-node.d.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    requisition: {
+      getRequisitionById(requisitionId: string): Promise<Requisition>;
+    };
+
+    institution: {
+      getInstitutions(options?: {country?: string}): Promise<Institution[]>;
+    };
+  }
+}
```
frdrkolsson: No newline

src/server/api/routers/nordigen-node.d.ts
```diff
@@ -0,0 +1,83 @@
...
+    bic: string
+    transaction_total_days: string
+    countries: string[]
+    logo: string
+  }
+
+  export default class NordigenClient {
+    constructor(options: {
+      secretId: string | undefined;
+      secretKey: string | undefined;
```
frdrkolsson: Thoughts on 
```suggestion
      secretId?: string;
      secretKey?: string;
```

src/server/api/routers/gocardless.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      const institutionsData = await client.institution.getInstitutions(country ? {country} : undefined)
+
+      if (!institutionsData) {
+        throw new Error('Failed to fetch institutions');
+      }
+  
+      return institutionsData;
+  }),
+})
```
frdrkolsson: No newline

src/server/api/routers/gocardless.ts
```diff
@@ -0,0 +1,103 @@
...
+      if (!init) {
+        throw new Error('Failed to link to bank');
+      }
+
+      return init;
+  }),
+  getInstitutions: protectedProcedure
+    .input(z.string().optional())
+    .query(async ({ input }) => {
+      const country = input
```
frdrkolsson:
```suggestion
    .query(async ({ input: country }) => {
```

src/server/api/routers/gocardless.ts
```diff
@@ -0,0 +1,103 @@
...
+        update: { data: JSON.stringify(transactions), lastFetched: new Date(), userId: ctx.session.user.id },
+        create: { gocardlessId: accountId ?? "", data: JSON.stringify(transactions), lastFetched: new Date(), userId: ctx.session.user.id },
+      });
+  
+      return transactions;
+  }),
+  connectToBank: protectedProcedure
+    .input(z.string().optional())
+    .mutation(async ({ input, ctx }) => {
+      const institutionId = input
```
frdrkolsson:
```suggestion
    .mutation(async ({ input: institutionId, ctx }) => {
```

src/server/api/routers/gocardless.ts
```diff
@@ -0,0 +1,103 @@
...
+const client = new NordigenClient({
+  secretId: env.GOCARDLESS_SECRET_ID,
+  secretKey: env.GOCARDLESS_SECRET_KEY
+});
+
+export const gocardlessRouter = createTRPCRouter({
+  getTransactions: protectedProcedure
+    .input(z.string().optional())
+    .query(async ({ input, ctx }) => {
+      const requisitionId = input
```
frdrkolsson:
```suggestion
    .query(async ({ input: requisitionId, ctx }) => {
```

alexanderwassbjer:
> this looks good overall with some comments. please lemme know what you think!
> 
> 
> 
> tysm for this PR really appreciate it. this is really awesome!
> 
> 
> 
> think it's best to merge in v2, test thoroughly and release it with other features! (i'll create a new branch)

Awesome! Iam happy you liked it.
I will go over it again, check your comments and fix some other stuff.

It would be nice if it was released in v2!

.env.example
alexanderwassbjer: Alright! Good catch, I will change it and have it returned from the server instead.
As you see in one of my dockerfiles, I had to put the public env in the build phase to get it to work. But it did not work in the best way ;)

.env.example
alexanderwassbjer: No, it should work world wide. The bank provider picker in the settings page fetch all bank providers. It works with either one country or undefined which then returns all the bank providers.
But because the list can be long in that case, I added the env to have a option to only fetch some.

If you have another approach to this, feel free to tell me.
One way is to have a country picker first and then the providers but it feels a bit weird..

Dockerfile
alexanderwassbjer: Good catch, will remove it!

docker-compose.yml
alexanderwassbjer: Good catch, will remove it!

docker/Dockerfile
alexanderwassbjer: I will it to a server env instead as you suggested in other comment :)

src/pages/_app.tsx
alexanderwassbjer: Of course! I can revert it and then make another pr for that!
A setting in settings page for either "light, dark or system prefer" would be the best option!
I will take a look at it

.env.example
KMKoushik: we have default currency option, we can get the country from that currency!

src/pages/_app.tsx
KMKoushik: legend thanks!

KMKoushik:
> 
> Awesome! Iam happy you liked it. I will go over it again, check your comments and fix some other stuff.
> 
> It would be nice if it was released in v2!

ofcourse this is going to be star feature!

.env.example
alexanderwassbjer: I can't find the default currency? If you mean the currency that is set by the user when they try to add a expense it feels a bit weird to use that because it can change?

alexanderwassbjer: So for now, it only works with one country in the package I use. https://github.com/nordigen/nordigen-node/blob/9d535ca9d788115d4e199c749432733be3ea6cdd/lib/api/institutions.js#L23

That can be fixed, either in the package or that we have a country picker first.
But if we think of the user case, will it be necessary to have multiple countries at the same instance?

src/components/AddExpense/GoCardlessTransactions.tsx
alexanderwassbjer: Take a new look :)

Dockerfile
alexanderwassbjer: https://github.com/oss-apps/split-pro/pull/132/commits/a1caeed1f7cdf5c4546b2ff8868fa528abf0f781

docker-compose.yml
alexanderwassbjer: https://github.com/oss-apps/split-pro/pull/132/commits/a1caeed1f7cdf5c4546b2ff8868fa528abf0f781

docker/Dockerfile
alexanderwassbjer: https://github.com/oss-apps/split-pro/pull/132/commits/25fc9faeefd3aadd0b30c1b626ccec66c3506bcb

.env.example
alexanderwassbjer: https://github.com/oss-apps/split-pro/pull/132/commits/25fc9faeefd3aadd0b30c1b626ccec66c3506bcb

src/pages/_app.tsx
alexanderwassbjer: Here is a new PR: https://github.com/oss-apps/split-pro/pull/133

.env.example
alexanderwassbjer: If we want to stick with the env and use it either with no value to fetch all bank providers or filter it based on the env. I have created a PR on the Nordigen-node package and hoping for a merge so we can use it.
https://github.com/nordigen/nordigen-node/pull/63

alexanderwassbjer:
@alexanderwassbjer
Some other ideas for improvement:
- Encrypt the cache for the transactions.
- Add multiple bank accounts for different providers.

KMKoushik:
i'll have to think about this over the weekend!

alexanderwassbjer:
@KMKoushik If this is hard to solve or if its because we are limited to the free tier of gocardless, maybe we can have so its for now only possible to use it if you self host?

Silther:
> It would be preferred to replace gocardless with Plaid.

Plaid supports almost no European banks, so you would have to add both.

alexanderwassbjer:
@gravelfreeman @Silther I think it’s better to start with gocardless, otherwise the PR will just sit there. Adding more providers like plaid in the future isn’t that complicated. This is mainly for those who are self-hosting. Finding a completely free solution that works for all users is hard.

alexanderwassbjer:
@krokosik What do you think?

krokosik:
I agree with the approach of adding one provider first (gocardless since it is already implemented) and then someone can add others. I'm sorry I did not look at this PR so far, but have been busy with more core issues since I took over as maintainer.

Recently I have been on vacation a lot and did not finish release 1.5, which I also hope to adjust now to finish up the core issues in my opinion:
- full splitwise import
- splitpro import for migrations
- separating total and friend balances
- minor but annoying UX issues

Once 1.5 is released you can rebase your PR and I can review it then.

alexanderwassbjer:
> I agree with the approach of adding one provider first (gocardless since it is already implemented) and then someone can add others. I'm sorry I did not look at this PR so far, but have been busy with more core issues since I took over as maintainer.
> 
> Recently I have been on vacation a lot and did not finish release 1.5, which I also hope to adjust now to finish up the core issues in my opinion:
> 
> * full splitwise import
> * splitpro import for migrations
> * separating total and friend balances
> * minor but annoying UX issues
> 
> Once 1.5 is released you can rebase your PR and I can review it then.

@krokosik Sounds good, and I really like your approach! I’ll finalize things once 1.5 is released. Please just let me know if you need any help along the way.

krokosik:
Great! Thank you for not losing interest in this feature, I know it's been a while.

In the meantime you can help with some 1.5 milestone features :)

alexanderwassbjer:
@krokosik Of course, I think it's a really helpful feature.

Do you have a detailed list of tasks for the 1.5 that I can check? :)

alexanderwassbjer:
@alexanderwassbjer Note for my self: "Send all" should have isLoading

alexanderwassbjer:
@krokosik I have made this PR up to date with the latest code in main and it's working!
And I also did some improvements to the code so it's easier to add more banking providers in the future 😊

alexanderwassbjer:
@krokosik Its ready for review.

Would also like someone to look at the translations for the other languages.

.env.example
```diff
@@ -46,6 +46,12 @@ EMAIL_SERVER_PORT=
 EMAIL_SERVER_USER=
 EMAIL_SERVER_PASSWORD=
 
+# GoCardless options
+GOCARDLESS_ENABLED=
```
krokosik: This env var is a bit unnecessary imo. We can just check if the required env vars below are provided, similiar to how we handle other optional services.

public/locales/pl/account_page.json
```diff
@@ -66,6 +66,13 @@
     "support_us": "Wesprzyj nas",
     "follow_on_x": "Śledź nas na X",
     "logout": "Wyloguj",
+    "choose_bank_provider": "Wybierz bank",
+    "select_bank_provider": "Wybierz bank",
+    "search_bank": "Wyszukaj bank",
+    "no_bank_providers_found": "Nie znaleziono banków",
+    "to_bank": "do banku",
+    "reconnect": "Znowu połącz",
```
krokosik: Połącz ponownie sounds better

src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -262,7 +391,19 @@ export const AddOrEditExpensePage: React.FC<{
...
+              </Button>
+            </div>
           </div>
+          <BankingTransactions
+            add={addViaBankTransaction}
+            addMultipleExpenses={addMultipleExpenses}
+            multipleTransactions={multipleTransactions}
+            setMultipleTransactions={handleSetMultipleTransactions}
+            isTransactionLoading={isTransactionLoading}
+          />
```
krokosik: This is rather bulky (with all the associated methods). And `AddExpensePage` is already huge and I plan to modularize it over time. Can you move it to a separate file?

src/components/Account/BankAccountSelect.tsx
```diff
@@ -0,0 +1,99 @@
...
+        </Button>
+      }
+      onTriggerClick={handleTriggerClick}
+      title={t('ui.select_bank_provider')}
+      className="h-[70vh]"
+      shouldCloseOnAction
+      open={open}
+      onOpenChange={handleOpenChange}
+    >
+      <div>
```
krokosik: What is the puprose of this empty div?

src/components/Account/BankAccountSelect.tsx
```diff
@@ -0,0 +1,99 @@
...
+          </div>
+          <ChevronRight className="h-6 w-6 text-gray-500" />
+        </Button>
+      }
+      onTriggerClick={handleTriggerClick}
+      title={t('ui.select_bank_provider')}
+      className="h-[70vh]"
+      shouldCloseOnAction
+      open={open}
+      onOpenChange={handleOpenChange}
```
krokosik: It seems to me there is no need to make it a controlled component, since you don't do any side effects on open close.

src/components/Account/BankAccountSelect.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const { t } = useTranslationWithUtils(['account_page']);
+
+  const [open, setOpen] = React.useState(false);
+  const userQuery = api.user.me.useQuery();
+  const updateProfile = api.user.updateUserDetail.useMutation();
+  const institutions = api.gocardless.getInstitutions.useQuery();
+  const gocardlessEnabled = api.gocardless.gocardlessEnabled.useQuery();
+
+  if (!gocardlessEnabled) {
+    return <></>;
```
krokosik: Return `null` instead:
https://legacy.reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

krokosik: Also, I would like to make sure that you've checked the file with oxc and made sure you do not introduce any new linter warnings.

src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -79,7 +88,89 @@ export const AddOrEditExpensePage: React.FC<{
...
+      }
+      return Promise.resolve();
+    });
+
+    await Promise.all(expensePromises);
+
+    setMultipleTransactions([]);
+    setIsTransactionLoading(false);
+    router
+      .push(group?.id ? `/groups/${group?.id}` : '/balances')
```
krokosik: Use `router.back`

src/components/Layout/MainLayout.tsx
```diff
@@ -77,10 +77,13 @@ const MainLayout: React.FC<MainLayoutProps> = ({
...
         </nav>
-        <div className="w-full overflow-auto lg:border-x lg:border-gray-900 lg:px-6">
+        <div
+          className="w-full overflow-auto lg:border-x lg:border-gray-200 lg:px-6 dark:lg:border-gray-900"
+          id="mainlayout"
+        >
           {title ? (
             <div className="mb-2 flex items-center justify-between px-4 py-4">
-              <div className="text-3xl font-bold text-gray-200">{title}</div>
+              <div className="text-3xl font-bold text-black dark:text-gray-200">{title}</div>
```
krokosik: What's the reason of changing the styling of almost all the views in the app?

src/lib/currency.ts
```diff
@@ -835,3 +835,7 @@ export const CURRENCIES = {
 };
 
 export type CurrencyCode = keyof typeof CURRENCIES;
+
+export function isCurrencyCode(value: string): value is CurrencyCode {
```
krokosik: Nice!

src/pages/account.tsx
```diff
@@ -92,6 +97,13 @@ const AccountPage: NextPageWithUser<{ isCloud: boolean; feedBackPossible: boolea
     void router.push('/auth/signin', '/auth/signin', { locale: 'default' });
   }, [router]);
 
+  const onConnectToBank = async () => {
+    const res = await connectToBank.mutateAsync(userQuery.data?.bankingId);
+    if (res?.link) {
+      window.location.href = res.link;
+    }
+  };
+
```
krokosik: Wrap with useCallback

src/pages/balances.tsx
```diff
@@ -50,7 +50,7 @@ const BalancePage: NextPageWithUser = () => {
         loading={balanceQuery.isPending}
       >
         <NotificationModal />
-        <div className="">
+        <div>
```
krokosik: I guess this empty div can be removed

src/server/api/routers/gocardless.ts
```diff
@@ -0,0 +1,113 @@
...
+          if (!cachedData.data) {
+            throw new Error('Failed to fetch cached transactions');
+          }
+          return JSON.parse(cachedData.data) as GetTransactions;
+        }
+      }
+
+      const account = client.account(accountId ?? '');
+
+      const transactions = await account.getTransactions();
```
krokosik: Do we fetch **ALL** the bank transactions here? Maybe we should only fetch transaction from the date selected in the edited expense?

src/server/api/routers/gocardless.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    if (!institutionsData) {
+      throw new Error('Failed to fetch institutions');
+    }
+
+    return institutionsData;
+  }),
+  gocardlessEnabled: publicProcedure.query(async () => {
+    return env.GOCARDLESS_ENABLED ?? false;
+  }),
```
krokosik: Making a trpc query of it is rather overkill. I would suggest passing this via `getServerSideProps`

src/server/api/routers/nordigen-node.d.ts
```diff
@@ -0,0 +1,80 @@
+declare module 'nordigen-node' {
```
krokosik, line 1: I checked on npm and it says `nordigen-node` is written in Typescript, why do we need a `d.ts` file?

src/components/Account/BankAccountSelect.tsx
alexanderwassbjer: Fair enough. But I actually close it upon bank select. But probably its fine to do it uncontrolled and skip the close on select

src/components/Account/BankAccountSelect.tsx
alexanderwassbjer: Think I had some style there before, will remove it! 👍

.env.example
alexanderwassbjer: Good point! Will change that

src/server/api/routers/nordigen-node.d.ts
alexanderwassbjer: Will have a look at that, before I was sure that the types was not correct and therefore I added this type file, but I agree with the package are typed

src/server/api/routers/gocardless.ts
alexanderwassbjer: Good point!

src/components/Account/BankAccountSelect.tsx
alexanderwassbjer: Good catch! I have also checked the files with the linter 👍

src/pages/account.tsx
alexanderwassbjer: 👍

src/pages/balances.tsx
alexanderwassbjer: 👍

src/server/api/routers/gocardless.ts
alexanderwassbjer: I’ll do that, starting by 90 days back. It could also be useful to have a filter on the transaction screen in the future, so the user can choose which dates to fetch

src/components/AddExpense/AddExpensePage.tsx
alexanderwassbjer: 👍

src/server/api/routers/nordigen-node.d.ts
alexanderwassbjer: I’ve tested it. The package does include types, but they aren’t correctly bundled with the build, which means the types aren’t properly exposed in the final output. That’s why I had to create my own declaration.

src/components/AddExpense/AddExpensePage.tsx
alexanderwassbjer: 👍

src/server/api/routers/nordigen-node.d.ts
krokosik: Indeed: https://github.com/nordigen/nordigen-node/issues/56

alexanderwassbjer: Exactly! Right now, though, the declared file sits inline with the package types. Hopefully that’ll get sorted out soon

src/components/Layout/MainLayout.tsx
alexanderwassbjer: That is actually because this PR was based on a v2 branch that KMKoushik did where we had merged a dark/light mode feature.
https://github.com/oss-apps/split-pro/pull/133

Probably it would be best to restore the changes of the styles in this PR and then maybe implement dark/light mode again in another PR.

alexanderwassbjer:
> I left some comments as an initial review from just looking at the code. For a more in depth one, I will need to get GoCardless myself.
> 
> I am also concerned about the cache never being deleted from the db. Something like `pg-cron` will be required. It would be great if you could also create some Typescript API for DB caching, as it would also come in handy with currency rates.
> 
> Finally, I would still hold on resolving conflicts until `1.5.0` stable is released. For example, we merged a Portugese locale now, which also needs to be updated with new keys and there may be more bug fixes on the way.

Thanks for your review! Good points. Best would be to signup for GoCardless.

I can take a look at implementing some cron inorder to remove the cache after like a month after its created / updated date. It's a good point.

I will make the changes, then wait for the 1.5 stable release and then resolve the conflicts 😊

src/server/api/routers/nordigen-node.d.ts
krokosik: I've sent them a quick PR: https://github.com/nordigen/nordigen-node/pull/66

Let's see if they do sth about it. If not, we can use my fork (via Git) or the Typescript rewrite that someone posted.

krokosik:
1.5 stable is out :)

Regarding cron there are 2 options that I consider. One is the scheduler in `pg-boss`, which I would like to add for running background tasks in a separate node process. It does have cron, but I haven't read into the tradeoffs of using it versus `pg-cron`.

The cache policy should be configurable, one month seems like a lot of transactions to me.

alexanderwassbjer:
> 1.5 stable is out :)
> 
> Regarding cron there are 2 options that I consider. One is the scheduler in `pg-boss`, which I would like to add for running background tasks in a separate node process. It does have cron, but I haven't read into the tradeoffs of using it versus `pg-cron`.
> 
> The cache policy should be configurable, one month seems like a lot of transactions to me.

Awesome with 1.5! Then we will try to merge this PR.

I haven’t read much about pg-boss, but I’m considering either that or pg-cron. My gut feeling is that having the cron jobs as a database extension with pg-cron seems more stable than handling them in code with pg-boss — though I might be wrong.

EDIT: The cache policy for manage is 1 day. But 1 month is the getTransactions from GoCardless interval (dateFrom, dateTo). Maybe that should be configurable?

src/server/api/routers/nordigen-node.d.ts
alexanderwassbjer: I think it would be good to start using your fork

alexanderwassbjer:
@krokosik I have pushed some new code where all your comments are fixed and some other stuff.

Whats left to do is:
- Use your fork of the nordigen-node
- Implement some cron for pg to remove cache older then like 2 days.

I would like your thoughts on the cron stuff.

.env.example
krokosik: Can you submit this PR to my fork? :)

krokosik:
Regarding cron, I will read more about it next week, but so far, it is unclear to me how well `pg-cron` integrates with Prisma. My initial lookup showed that Postgres extensions are supported as a preview feature, but I have no idea how well. I would also prefer to use `pg-cron` as a more battle tested and stable solution, but if it's a nightmare in Prisma (which is often the case with advanced DB features) we can skip that.

On the other hand we will use `pg-boss` anyway. Question is how extensively. If we could have both that would be great. If you could add `pg-cron` as a proof of concept that would certainly speed things up 🚀

.husky/pre-commit
```diff
@@ -1 +1,2 @@
 pnpm lint-staged
+pnpm tsc --noEmit
```
krokosik: Maybe a separate PR with that? I support adding this and we can benefit from it sooner :)

alexanderwassbjer:
> Regarding cron, I will read more about it next week, but so far, it is unclear to me how well `pg-cron` integrates with Prisma. My initial lookup showed that Postgres extensions are supported as a preview feature, but I have no idea how well. I would also prefer to use `pg-cron` as a more battle tested and stable solution, but if it's a nightmare in Prisma (which is often the case with advanced DB features) we can skip that.
> 
> On the other hand we will use `pg-boss` anyway. Question is how extensively. If we could have both that would be great. If you could add `pg-cron` as a proof of concept that would certainly speed things up 🚀

Indeed! I’ve seen that this is still a preview feature, and at the moment Prisma only supports a limited set of extensions.
https://www.prisma.io/docs/postgres/database/postgres-extensions
Quote:
image

Another option would be to manually install `pg_cron` through a migration, but that approach isn’t really recommended either. And we would have to install the pg_cron inside the psql container in docker in order to make the migration to run.
https://github.com/prisma/prisma/issues/12308

I will have a look at implementing `pg-boss` 👍 

EDIT: I'am not sure how our self-hosters is running their instance of split-pro. But Netlify, Vercel for example don't keep workers alive, like `pg-boss` or `node-cron` in a separate node instance. Instead they recommend using their own way of managing cron jobs.
The self-hosters that run in docker on a local server, my self included would probably not have any issues with `pg-boss` being killed.

.husky/pre-commit
alexanderwassbjer: :) https://github.com/oss-apps/split-pro/pull/350

alexanderwassbjer:
@krokosik I have added a proof of concept of the `pg_cron`. It works good without any problems. [3e873a0](https://github.com/oss-apps/split-pro/pull/132/commits/3e873a02f382a81dea0ccda85d21892a54b99bf8) I have some findings:
- Could not add `pg_cron` and schedule jobs in a prisma migration because it creates a shallow copy which do not have the `pg_cron` extension available.
- Created a cron setup file which can be called multiple times without duplicated cron jobs being added.
- Added a separate db:cron script which allows us to either run it after the db migration script or separate.
- Had to install `pg_cron` in the psql instance via a dockerfile which is used in the compose file.
- The `pg_cron` is logging all of its runs in a table called `cron.job_run_details`.
image

I have tested the implementation a bit and I think it's a good approach.

krokosik:
Awesome work! Will review the changes this week.

krokosik:
Oof bummer, gocardless does not support individual accounts in my country (Poland) and I don't have a business. We need to figure out a way to test it, are there some other providers I could try?

alexanderwassbjer:
> Oof bummer, gocardless does not support individual accounts in my country (Poland) and I don't have a business. We need to figure out a way to test it, are there some other providers I could try?

Ohno! But do you have like Revolut, because they are supported. Otherwise my plan is to maybe implement the plaid api aswell.

krokosik:
The requirement to own a business is a requirement for a gocardless account itself. I didn't event get to adding accounts.

alexanderwassbjer:
> The requirement to own a business is a requirement for a gocardless account itself. I didn't event get to adding accounts.

Hmm what?
If you go to: https://developer.gocardless.com/bank-account-data/overview

Then it says:
"Before you start, please make sure you have acquired a user secret from GoCardless's Bank Account Data portal"

There is a link you can follow.

Which is free and no requirement of a business. That's the way it went and it works good.

krokosik:
image

I can't even create an account, but in the developer page I noticed I can make a sandboxed account for testing, which seems perfect. I will be less hesitatnt to give fake info there if I don't have to connect my real bank account 🤞

krokosik:
Aaaand the sandbox account does not let me into the Bank Account Data portal.

krokosik:
Okay nevermind, they have a separate sandbox for bank account data! So I don't even need to make an account there <3

alexanderwassbjer:
> Okay nevermind, they have a separate sandbox for bank account data! So I don't even need to make an account there <3

Awesome! 🤩

alexanderwassbjer:
@krokosik I just had to verify that it worked to create an account without a business.
https://bankaccountdata.gocardless.com/login
Then signin with a google account which will create an account.
It worked for me twice with no problem.
😊

krokosik:
@alexanderwassbjer yeah, my bad. I didn't realize bank account data was a separate service. With that being said, I'm a bit swamped now with other responsibilities sadly.

krokosik:
Okay, I've finished off currency conversions and have a bit more time. I will see if I can use the dummy BankAccount data, as I still can't make an account. The website you linked is for login only and I cannot register an account there.

krokosik:
On another note, Postgres extensions are now stable in Prisma 6.16 :)

alexanderwassbjer:
> Okay, I've finished off currency conversions and have a bit more time. I will see if I can use the dummy BankAccount data, as I still can't make an account. The website you linked is for login only and I cannot register an account there.

Nice! If you try to login with google on the link I sent it works for me to register a new account.

krokosik:
Like I said, it doesn't work for me

alexanderwassbjer:
> Like I said, it doesn't work for me

Hmm weird! I will continue look at that so it's correct in the readme then.
Hope it works with the sandbox for you!

krokosik:
Okay, I will only pick up minor issues now and focus on merging this PR next. As such could you make a rebase/merge of recent changes?

alexanderwassbjer:
> Okay, I will only pick up minor issues now and focus on merging this PR next. As such could you make a rebase/merge of recent changes?

Awesome, thanks! I will fix it and rebase 👍

alexanderwassbjer:
@krokosik I have now done the rebase/merge so its up to date. Please look at it so we can try to merge it :)

alexanderwassbjer:
@krokosik I did some changes and because of my previous refactoring I was able to add plaid alongside with gocardless. Everything are the same functionality except the fetching of the data that are different with plaid.

krokosik:
@alexanderwassbjer it seems the sandbox bank account portal no longer exists. I also found this fairly recent thread: https://www.reddit.com/r/selfhosted/comments/1mfs6ro/actual_budget_cannot_sign_up_to_gocardless_bank/?rdt=57410

It seems they are in the process of removing this service, so I don't think this is a provider we want to use in a free OSS project.

krokosik:
Can you add some docs on setting up a Plaid account?

krokosik:
Oh for f*ck sake, I feel so done with all these providers after having wasted so much time on them now

image

`{"error":"","isRecaptchaFailed":false}`

Personally I don't plan on using this feature and I don't want to commit anymore time to creating accounts etc. If you have some way for me to test these features with a sandbox environment or something, then we can progress with the PR. Otherwise I don't see how I could maintain it in the future.

alexanderwassbjer:
> Oh for f*ck sake, I feel so done with all these providers after having wasted so much time on them now
> 
> 
> 
> image
> 
> 
> 
> `{"error":"","isRecaptchaFailed":false}`
> 
> 
> 
> Personally I don't plan on using this feature and I don't want to commit anymore time to creating accounts etc. If you have some way for me to test these features with a sandbox environment or something, then we can progress with the PR. Otherwise I don't see how I could maintain it in the future.

Haha shit! I will create a Readme and look at the gocardless future plans.
I have a sandbox I can share. Do you have like discord or something so I can share it in private to you?

krokosik:
@alexanderwassbjer awesome! Feel free to add me on Discord #_krokosik

.env.example
```diff
@@ -46,6 +46,23 @@ EMAIL_SERVER_PORT=
 EMAIL_SERVER_USER=
 EMAIL_SERVER_PASSWORD=
 
+# GoCardless options
+GOCARDLESS_COUNTRY=
+GOCARDLESS_SECRET_ID=
+GOCARDLESS_SECRET_KEY=
+# Bank Transactions will be fetched from today and 30 days back as default.
+GOCARDLESS_INTERVAL_IN_DAYS=
```
krokosik, line 54: Following our discussion, GoCardless will have to be removed as they closed of their service to new users

README.md
```diff
@@ -122,6 +122,7 @@ pnpm i
 - Copy the env.example file into .env
 - Setup google oauth required for auth https://next-auth.js.org/providers/google or Email provider by setting SMTP details
 - Login to minio console using `splitpro` user and password `password` and [create access keys](http://localhost:9001/access-keys/new-account) and the R2 related env variables
+- If you want to use bank integration please create a free account on [GoCardless](https://gocardless.com/bank-account-data/) or [Plaid](https://plaid.com) and then enter the the related env variables
```
krokosik: Remove GoCardless and add more detailed Plaid setup instructions in a separate MD file or a link to a tutorial (include screenshots)

docker/dev/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM postgres:16
```
krokosik: Remove in favor of `ossapps/postgres`

package.json
```diff
@@ -56,12 +57,15 @@
     "next-i18next": "^15.4.2",
     "next-themes": "^0.2.1",
     "nodemailer": "^6.9.8",
+    "nordigen-node": "^1.4.1",
```
krokosik, line 59: Remember to remove this dep along with gocardless

prisma/pg_cron_setup.sql
```diff
@@ -0,0 +1,23 @@
+CREATE EXTENSION IF NOT EXISTS pg_cron;
```
krokosik: I have created a solution at the `160-recurring-expense` branch that calls this stuff from the Nextjs app instead.

prisma/schema.prisma
```diff
@@ -66,6 +68,16 @@ model User {
     updatedExpenses     Expense[]            @relation("UpdatedByUser")
 }
 
+model CachedBankData {
+    id              Int      @id @default(autoincrement())
+    obapiProviderId String   @unique
+    data            String
+    userId          Int
```
krokosik, line 76: Let's add a relation to the User

public/locales/en/common.json
```diff
@@ -175,6 +175,20 @@
     "edit_name": {
       "title": "Edit name",
       "placeholder": "Enter name"
+    },
```
krokosik: Since v1.6 I plan to drop the requirement to update ALL translation when adding new features. We are in the process of setting up WebPlate, so that translations can be updated through a nice GUI and tracked online.

There are just way too many languages to maintain all of that and it is especially annoying during development. As such you can limit yourself to just English.

On another note, I would propose to put all these bank related keys into a common group in the json

public/locales/en/common.json
```diff
@@ -175,6 +175,20 @@
...
+    "choose_bank_provider": "Choose bank provider",
+    "select_bank_provider": "Select bank provider",
+    "search_bank": "Search bank",
+    "no_bank_providers_found": "No bank providers found",
+    "to_bank": "to bank",
+    "reconnect": "Reconnect",
+    "connect": "Connect",
```
krokosik: These string belong in the actions group

src/components/Layout/MainLayout.tsx
```diff
@@ -79,7 +79,10 @@ const MainLayout: React.FC<MainLayoutProps> = ({
             currentPath={currentPath}
           />
         </nav>
-        <div className="w-full overflow-auto lg:border-x lg:border-gray-900 lg:px-6">
+        <div
+          className="w-full overflow-auto lg:border-x lg:border-gray-900 lg:px-6"
+          id="mainlayout"
+        >
```
krokosik, line 85: Another 2.0 styling change?

src/pages/account.tsx
```diff
@@ -110,6 +116,11 @@ const AccountPage: NextPageWithUser<{ feedBackPossible: boolean }> = ({
             </AccountButton>
           </LanguagePicker>
 
+          <BankConnection
+            bankConnectionEnabled={bankConnectionEnabled}
+            bankConnection={bankConnection}
+          />
+
           {isCloud && (
```
krokosik, line 131: Please update to the new pattern, where the trigger button is declared here and passed in as a child component. The child component has to be an `AccountButton`

src/pages/balances.tsx
```diff
@@ -71,17 +69,15 @@ const BalancePage: NextPageWithUser = () => {
                       <span className="text-orange-600">
                         {balance.currency.toUpperCase()} {toUIString(balance.amount)}
                       </span>
-                      {index !== balanceQuery.data.youOwe.length - 1 ? (
-                        <span className="">+</span>
-                      ) : null}
+                      {index !== balanceQuery.data.youOwe.length - 1 ? <span>+</span> : null}
```
krokosik, line 72: It seems there are more 2.0 changes here? Please revert tot minimize the PR

src/server/api/routers/nordigen-node.d.ts
```diff
@@ -0,0 +1,86 @@
...
```
krokosik, line 1: This can be removed now, together with gocardless

src/server/api/services/splitService.ts
```diff
@@ -41,6 +41,7 @@ export async function createExpense(
     participants,
     expenseDate,
     fileKey,
+    transactionId,
```
krokosik, line 44: What about editing expenses?

src/server/bankTransactionHelper.ts
```diff
@@ -0,0 +1,45 @@
+import { z } from 'zod';
+import { env } from '~/env';
+
+type BankProviders = 'GOCARDLESS' | 'PLAID';
+
+export const InstitutionsOutput = z.array(
```
krokosik: The Zod schemas should go into `src/types/bank.types.ts`

src/server/bankTransactionsController/goCardless/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+  DEFAULT_LANGUAGE: 'EN',
+  DATE_FORMAT: 'yyyy-MM-dd',
+} as const;
+
+export const ERROR_MESSAGES = {
+  FAILED_FETCH_CACHED: 'Failed to fetch cached transactions',
+  FAILED_FETCH_INSTITUTIONS: 'Failed to fetch institutions',
+  FAILED_FETCH_TRANSACTIONS: 'Failed to fetch transactions',
+  FAILED_LINK_BANK: 'Failed to link to bank',
+} as const;
```
krokosik: These need to be localized I believe?

src/server/bankTransactionsController/goCardless/constants.ts
```diff
@@ -0,0 +1,13 @@
+export const GOCARDLESS_CONSTANTS = {
```
krokosik: Subject to removal as well right?

src/server/bankTransactionsController/plaid/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { format, subDays } from 'date-fns';
```
krokosik: I like the structuring and the attention to detail. Great job!

May I suggest, however, creating an abstract provider class in the same way that was done for the currency rate providers? That would make implementing new providers clearer, as well as allow the trpc router to be provider agnostic, as in move the check logic away from the file.

We could then also make a provider "factory" like for the currency rates.

Finally, I would stick to the `service` nomenclature, as to my knowledge a backend controller is responsible mostly for handling requests/responses and calling appropriate functionality services.

src/server/api/routers/bankTransactions.ts
```diff
@@ -0,0 +1,111 @@
...
+    .input(z.string().optional())
+    .output(TransactionOutput.optional())
+    .query(async ({ input: token, ctx }) => {
+      const provider = whichBankConnectionConfigured();
+      if (provider === 'GOCARDLESS') {
+        return await gocardless.getTransactions(ctx.session.user.id, token);
+      }
+      if (provider === 'PLAID') {
+        return await plaid.getTransactions(ctx.session.user.id, token);
+      }
```
krokosik: Do we want to fetch ALL the transactions on every request? The caching helps with calling the provider API, but this might be a lot of data to collect and frontend wise it might be difficult to navigate.

The least we could do is filter by date, where we only fetch expenses from the date input in the add screen. What do you think?

src/components/Expense/ExpenseDetails.tsx
```diff
@@ -63,6 +64,12 @@ const ExpenseDetails: React.FC<ExpenseDetailsProps> = ({ user, expense, storageP
                 {t('ui.on')} {toUIDate(expense.createdAt, { year: true })}
               </p>
             )}
+            {expense.transactionId && (
+              <p className="flex gap-2 text-sm text-gray-500">
+                {t('expense_details.verified_transaction')}
+                <BadgeCheck className="h-4 w-4 text-emerald-500" />
+              </p>
+            )}
```
krokosik: Verified transaction seems a bit misleading to me. Maybe something along the lines of `bank transaction`? The icon could also be changed to `landmark` or sth like that. Also, there are a lot of gray text lines here now, it would be nice to break it apart.

What do you think about just putting the landmark icon next to the expense title?

src/components/AddExpense/BankTransactions/BankTransactionItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <div className="flex items-center gap-4">
+        <Checkbox
+          checked={multipleTransactions?.some(
+            (cItem) => cItem.transactionId === item.transactionId,
+          )}
+          disabled={alreadyAdded}
+          onCheckedChange={createCheckboxHandler(item)}
+          className="h-6 w-6 md:h-4 md:w-4"
+        />
+        <Button className="flex items-center gap-4" variant="ghost" disabled={alreadyAdded}>
```
krokosik: This is not very intuitive to me. With such a list, especially on mobile you just see checkboxes and may not be aware that it is possible to fill in the expense details using just a single transaction.

On the other hand, given a list of things to select, I would expect that clicking a row will toggle the checkbox and it selects a single transaction instead.

I will take the opportunity to lay out in this comment my main user flow icks:
- the new UI is nice but it severely cluttters the previously minimal add expense UI
- it kind of behaves like a separate entity to the original flow, especially with multiple transactions. This is especially bad since it is not obvious that the split outlined above is actually applied
- when selecting multiple transactions, there is no way to see the total sum and the fact that there are multiple expenses created is surprising to me.
- the add container gets expanded and overflows the screen, putting either the sponsor button or the transaction selection

Here are some loose thoughts I have on next steps, which are up to debate of course:
- the transaction selection should stay out of the user's way - I would propose putting it into an `AppDrawer` triggered by a `landmark` icon to the right of the sponsor button
- inside of it, the user can select multiple transactions. You may want to consider the scroll-area component from shadcn, but maybe the appdrawer handles that already
- regarding multiple transactions, in the current flow, you cannot modify them individually, which implies to me that you are creating one expense out of multiple transactions (for example buying groceries in several stores). So instead of creating multiple expenses, I would just update the add page with the total sum AND a title that concatenates all the transaction name. As I doubt it will be a sensible name, we can focus the title input after closing the modal.

Let me know what do you think about it.

krokosik (changes requested):

.env.example
alexanderwassbjer: For everyone else information: @krokosik and I has decided to keep gocardless for now but add a deprecated flag on it.

src/server/api/routers/bankTransactions.ts
alexanderwassbjer: I do indeed fetch only 30 days back. Because the filters work differently from gocardless and plaid that implementation is in the "controllers".

I agree that we should in version 2 of this add filter by date or like search etc.

src/pages/balances.tsx
alexanderwassbjer: No, that is just a lint fix where there was a empty className

README.md
alexanderwassbjer: Have added a separate readme for it

public/locales/en/common.json
alexanderwassbjer: Awesome! Looking forward to the new translation gui so new feature not misses translations.

I have moved into a own group!

src/components/Layout/MainLayout.tsx
alexanderwassbjer: No, this is for a small fix for the responsiveness in mobile when having it as a PWA app. Maybe suited for a own PR but I think it's too small for that

src/pages/account.tsx
alexanderwassbjer: Done!

src/server/api/services/splitService.ts
alexanderwassbjer: Editing expenses still work, the transactionid will remain the same regarding to the code. Then maybe we can think of when someone changes a expense that are "verified" it should maybe remove the transactionid. But right now I think not like that.

src/server/bankTransactionsController/goCardless/constants.ts
alexanderwassbjer: The use are for trpcerror and I saw that the other routers that have trpcerrors are not localized which is the reason this is not localized

src/server/bankTransactionsController/plaid/utils.ts
alexanderwassbjer: Good point! Have changed it

src/components/Expense/ExpenseDetails.tsx
alexanderwassbjer: Verified transaction in this meaning is that the other participants can verify that the expense actually are a real transaction. That's why it's there.

I have moved it into the title and with the landmark icon instead

src/components/AddExpense/BankTransactions/BankTransactionItem.tsx
alexanderwassbjer: I’ve updated the flow so that transaction selection is now moved out of the main form — there’s a button next to the Sponsor button which opens an AppDrawer. Inside the drawer you can select one or multiple transactions (either by clicking the checkbox or the entire row). When you’re done, there’s a submit button that adds them. This should make the main add expense view feel less cluttered and more focused.

But I actually think that the transaction button next to the sponsor button does not feel like a function button. I think it should be redesigned so it easier to understand that it's a feature.

Regarding the multiple transactions flow: I don’t agree with the idea of merging them into a single expense. In my implementation, each selected transaction is added as its own expense. The reason is that it allows you to later review, edit, or remove individual expenses. If we instead created a combined expense, that granularity would be lost.

From my perspective (and the way I use this in my own household finances with my partner), the ability to keep the transactions separate is essential. For example, if we mistakenly add the wrong transaction, it’s very easy to spot and remove the specific one. If everything were merged into a single expense, we would no longer use the feature since we couldn’t make such corrections without removing the whole combined expense.

I understand your point about the total sum not being visible at the moment — that’s a valid concern, and I think we could address it by showing a summary of the selected transactions somewhere in the AppDrawer before submitting in either this version of the feature or in v2 of it.

src/components/Expense/ExpenseDetails.tsx
alexanderwassbjer: image

src/components/AddExpense/BankTransactions/BankTransactionItem.tsx
alexanderwassbjer: image
image
image

src/server/api/services/dbCache.ts
```diff
@@ -0,0 +1,56 @@
...
+}: GetDbCachedDataParams<K>): Promise<T | null> {
+  const cached = await db[key].findUnique({
+    where: where,
+  });
+
+  if (!cached) {
+    return null;
+  }
+
+  const isFresh = cached.lastFetched > new Date(Date.now() - maxAgeMs);
```
krokosik: Can't we do that in DB?

src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -22,12 +22,14 @@ import { UploadFile } from './UploadFile';
 import { UserInput } from './UserInput';
```
krokosik, line 26: There seems to be a regression in AddExpense page causing runtimer errors on refresh. To reproduce:
- go to a group > add expense
- type some stuff in but don't submit
- refresh

Image

src/components/Account/BankAccount/PlaidLink.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const onPlaidSuccess = useCallback(
+    async (publicToken: string, _metadata: any) => {
+      setIsLoading(true);
+      try {
+        await exchangePublicToken.mutateAsync(publicToken);
+        toast.success(t('account.plaid.bank_connected_successfully'));
+        onSuccess?.();
+      } catch (error) {
+        console.error('Error exchanging public token:', error);
+        toast.error(t('account.plaid.bank_connection_failed'));
```
krokosik: It seems that the translation keys were not updated in the code after they were regrouped.

src/components/AddExpense/BankTransactions/BankingTransactionList.tsx
```diff
@@ -0,0 +1,148 @@
...
+          <Button
+            variant="outline"
+            className="text-md hover:text-foreground/80 justify-between rounded-full border-teal-500"
+          >
+            <div className="flex items-center gap-4">
+              <Landmark className="h-5 w-5 text-teal-500" />
+              {t('expense_details.bank_transactions')}
+            </div>
+          </Button>
+        </div>
```
krokosik: This trigger should also be passed as `children`. Currently it obscures styling in the add expense page component and you need to drill through 2 components

src/components/AddExpense/BankTransactions/BankTransactionItem.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { useCallback } from 'react';
+import { Button } from '../../ui/button';
+import { Checkbox } from '../../ui/checkbox';
+import type { TransactionAddInputModel } from '~/types';
+import { useTranslationWithUtils } from '~/hooks/useTranslationWithUtils';
+import { cn } from '~/lib/utils';
+import type { TransactionWithPendingStatus } from './BankingTransactionList';
+
+export const BankTransactionItem: React.FC<{
```
krokosik, line 9: Some transactions could be positive (like returns) and we do allow negative expenses (incomes). However, the item always expresses items in green. I would propose changing to red expenses, green incomes and grey already added transactions

Image

src/components/AddExpense/BankTransactions/BankingTransactionList.tsx
```diff
@@ -0,0 +1,148 @@
...
+      }
+      title={t('expense_details.bank_transactions')}
+      open={open}
+      onOpenChange={setOpen}
+      className="h-[80vh]"
+      actionTitle={t('expense_details.submit_all')}
```
krokosik: It seems we lost the ability to fill in the form with a single expense. I can only submit multiple.

The design is nice, but I would retain the ability to select a transaction and edit it afterwards.

src/components/AddExpense/BankTransactions/BankTransactionItem.tsx
krokosik: My proposal for the trigger button design:
image


On top of that, it should turn green when a transaction is selected and display an X to clear the selected transition, instead of the current clear button.

krokosik: Regarding multiple transactions:
I agree that creating multiple separate expenses would be nice, but this is the component for adding a single expense and I believe it should stay that way for consistent user experience. In my opinion, if you want granularity, you should just add them one by one, since the way you described, you still need to go over each individual expense.

I strongly feel that this page should **primarily** support adding a single expense. The button is now at a very convenient position for a finger to do it fast and the added convenience, especially that cancelling its consequences when done by mistake (very easy now) is quite tedious, should be possible in some other way.

Perhaps we could have a checkbox/dialog asking whether we want to merge the selected transactions into the edit form or submit seperately. The latter should show some clear information that it is going to happen (multiple expenses with the current split settings will be created). Another issue is that users can submit the expenses even without getting to the split type selection.

All in all, I just can't accept the multiple transaction in the current state and I don't have a good idea for its implementation. I propose first merging a singular transactions as the PR is vast and then we can iterate once we get a good idea on how to implement that.

prisma/migrations/20250922183043_add_cached_bank_data_relation/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AddForeignKey
+ALTER TABLE "public"."CachedBankData" ADD CONSTRAINT "CachedBankData_userId_fkey" FOREIGN KEY ("userId") REFERENCES "public"."User"("id") ON DELETE CASCADE ON UPDATE CASCADE;
```
krokosik: We can merge this into a single migration

src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -81,6 +87,7 @@ export const AddOrEditExpensePage: React.FC<{
   );
 
   const addExpense = useCallback(async () => {
+    const { group, paidBy } = useAddExpenseStore.getState();
```
krokosik: Why is it called inside of the hook? Why not specify it among other state values above and put them in the dependency array?

src/components/AddExpense/BankTransactions/BankTransactionItem.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  // const createCheckboxHandler = useCallback(
+  //   (item: TransactionWithPendingStatus) => () => onTransactionRowClick(item, true),
+  //   [onTransactionRowClick],
+  // );
+
+  const createClickHandler = useCallback(
+    (item: TransactionWithPendingStatus) => () => onTransactionRowClick(item, false),
+    [onTransactionRowClick],
+  );
```
krokosik, line 27: This will not be properly memoized I believe, as you return a function on each call anyway. It should rather look like:
```typescript
const createClickHandler = useCallback(
  () => onTransactionRowClick(item, false),
  [onTransactionRowClick, item],
);

File: src/components/AddExpense/AddExpensePage.tsx [hunk 59]
```diff
@@ -281,7 +303,29 @@ export const AddOrEditExpensePage: React.FC<{
...
           </div>
+          <AddBankTransactions
+            // clearFields={clearFields}
+            onUpdateAmount={onUpdateAmount}
+            bankConnectionEnabled={bankConnectionEnabled}
+          >
+            <div className="flex w-full justify-center">
+              <Button
+                variant="outline"
+                className="text-md hover:text-foreground/80 justify-between rounded-full border-teal-500"
```
krokosik: It seems the UI changes we discussed were not made?

.env.example
```diff
@@ -63,6 +63,9 @@ PLAID_COUNTRY_CODES=
 # Bank Transactions will be fetched from today and 30 days back as default.
 PLAID_INTERVAL_IN_DAYS=
 
+# Cron-job options
+CRON_JOB_FREQUENCY=
```
krokosik: Clear cache frequency? That name is rather vague considering we will have other cron jobs

src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -281,8 +302,28 @@ export const AddOrEditExpensePage: React.FC<{
                 </div>
               </>
             ) : null}
+            <div className="flex items-center justify-end gap-4">
+              <Button variant="ghost" className="text-primary px-0" onClick={clearFields}>
+                {t('expense_details.clear')}
+              </Button>
+            </div>
```
krokosik: This button should be removed as well. We discussed the UI changes and that clearing should be done via an icon button next to the transaction trigger. The transaction icon (landmark) should also turn to primary or green color upon selecting a transaction.

src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -281,8 +302,28 @@ export const AddOrEditExpensePage: React.FC<{
...
+          </div>
+          <div className="absolute right-0 bottom-30 left-0 flex items-center justify-between gap-4 lg:relative lg:bottom-0">
```
krokosik: justify-around looks nicer imo

krokosik (changes requested):
More UI comments. Especially that we have a responsiveness regression. Here is the comparison with the main branch

https://github.com/user-attachments/assets/d407e83f-dc6d-4453-ad8e-17343aa49ab0

krokosik:
@alexanderwassbjer thank you for all the hard work! I think that feature wise the branch is now ready to merge and we will polish it out+add multiple transactions as other developments in 1.6! Once the CI checks pass, we can go ahead and finally merge this.

Issue #138, opened 2024-10-29
Improvement: Order the groups by latest expense added


alexanderwassbjer:
It would be nice if the groups in the groups page were ordered by the latest expense added to the group. So the 1 group in the list would be the group that had the most recent expense added to it.

PR #139, opened 2024-10-29
🚀 Order by expense created date in the groups list

alexanderwassbjer:
Orders the groups on the group page so its based on the groups latest expense.
Before the groups were rendered in the default orderby which was order by id. Now it's ordered by the groups latest expense.

Related to this: https://github.com/oss-apps/split-pro/issues/138


alexanderwassbjer:
@KMKoushik Ready to be reviewed.

KMKoushik (approved):
this is great thanks

PR #141, opened 2024-10-29
🚀 Add group created date and first expense date

alexanderwassbjer:
- Added the group created date to group info panel.
- Added the groups first expense date in the groups statistics panel.

Screenshots:

image
image

### DIFF
--- a/src/pages/groups/[groupId].tsx
+++ b/src/pages/groups/[groupId].tsx
@@ -163,6 +163,17 @@ const BalancePage: NextPageWithUser<{
                     ) : null;
                   })}
                 </div>
+                {expensesQuery?.data && expensesQuery?.data?.length > 0 && (
+                  <div className="mt-8">
+                    <p className="font-semibold">First expense</p>
+                    <p>
+                      {format(
+                        expensesQuery.data[expensesQuery.data.length - 1]?.createdAt ?? new Date(),
+                        'yyyy-MM-dd',
+                      )}
+                    </p>
+                  </div>
+                )}
               </div>
             </AppDrawer>
             <AppDrawer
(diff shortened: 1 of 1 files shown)


alexanderwassbjer:
@KMKoushik Ready to be reviewed

src/pages/groups/[groupId].tsx
```diff
@@ -184,6 +195,12 @@ const BalancePage: NextPageWithUser<{
                   ))}
                 </div>
               </div>
+              {groupDetailQuery?.data?.createdAt && (
+                <div className="mt-8">
+                  <p className="font-semibold ">Group created</p>
+                  <p>{format(groupDetailQuery.data?.createdAt, 'yyyy-MM-dd')}</p>
```
Comment (KMKoushik) [inline src/pages/groups/[groupId].tsx]: `toLocaleDateString()` can we use this instead, so it will be formated according to user's locale?

KMKoushik (changes requested):
LGTM, just a small comment

src/pages/groups/[groupId].tsx
Comment (alexanderwassbjer) [reply src/pages/groups/[groupId].tsx]: Good idea! A note: when i18n should be implemented the format should also be localed.

Comment (alexanderwassbjer) [reply src/pages/groups/[groupId].tsx]: @KMKoushik Isn't that a problem for all the dates right now, because I cant see that toLocaleDateString is used anywhere in the project. In that case probably it should be added to all the places where dates are returned?

Comment (KMKoushik) [reply src/pages/groups/[groupId].tsx]: we can change for this and update others when we change those files, mostly we don't have to change cuz, it will be just month and date, for space constraint

Comment (alexanderwassbjer) [reply src/pages/groups/[groupId].tsx]: Fixed, I kept the format function as a wrapper just in case we want to change it like to MMM dd or something else.

KMKoushik (approved):

PR #152, opened 2024-12-11
Add authentik provider

batizisanya:
Add authentik provider support based on https://next-auth.js.org/providers/authentik


KMKoushik (approved):
not a big deal but will be nice if an svg is added in signin page, going to merge this for npw

idoodler:
Cool, but whats the redirect URL I have to configure in authentik?

PR #153, opened 2024-12-11
Add nextauth internal url

batizisanya:
Add option to set NEXTAUTH_URL_INTERNAL for nextauth. 
It's useful in environments when the server doesn't have access to the canonical URL of the site. (NEXTAUTH_URL)


KMKoushik (changes requested):
can you also add this compose file?

batizisanya:
> can you also add this compose file?

In general, I don't think it's necessary to add the NEXTAUTH_URL_INTERNAL variable to the docker compose file, because that would just lead to misunderstandings by having to set the same url twice. If I add this and the user forget to edit, then the authentication will not work. 

For those users who are in the rare situation that they need to use this setting, maybe I would prefer to provide guidance on this in the documentation, but I'm open to your opinion on this.

KMKoushik (approved):
LGTM

PR #175, opened 2025-01-26
feat: invite-only mode

### LINKED ISSUE #172: Feature Request: Invite-Only Mode
As far as I can tell this isn't currently supported.

It'd be helpful to have the ability to restrict sign-ups after the initial user has signed up on self-hosted installations.

As a self-hosted administrator of an instance, I want to support magic links but not allow ANYONE to sign up so I can avoid spam users.

---

This is something I'll be working on as long as there's no qualms with it getting merged in.

imnotjames:
this adds the env var `INVITE_ONLY` which forces the application into an invite-only mode:

Only users that have been invited by an existing user will be able to log in to the application.
Any other user will be rejected.

This does not stop the auth from being attempted & because the application doesn't have error handling on the sign in page users may not realize what's happening.

fixes #172


imnotjames:
probably need to add a message to the login page about this

KMKoushik (approved):

ldehner:
@imnotjames @KMKoushik, i set these to envs, but can still signup with new users
```shell
-e 'INVITE_ONLY'='true'
-e 'ENABLE_SENDING_INVITES'='true'
```
any ideas?

ldehner:
@imnotjames @KMKoushik i set INVITE_ONLY=true but still get a valid magic link to unregistered emails

imnotjames:
> @imnotjames @KMKoushik i set INVITE_ONLY=true but still get a valid magic link to unregistered emails

You ran off of main?

ldehner:
@imnotjames I am running ossapps/splitpro:latest (Docker on Unraid)

Soulplayer:
How can I run this on docker if it is not the latest version?

imnotjames:
> How can I run this on docker if it is not the latest version?

You can build your own image or run without docker

Soulplayer:
I'll wait then until the next release... :)

laur89:
dockerhub has 1.5.3, yet setting INVITE_ONLY=true still shows the email form & OTP is sent to mail. Is this expected?

PR #197, opened 2025-02-17
Simplify debts + group balances

### LINKED ISSUE #58: Option to simplify Debts

### LINKED ISSUE #158: Feature Request: Group Balance Page and Cash Flow Minimization
Currently we cannot see the full summary of who owes whom and what amount for the groups. This would be nice because most of the time one person tries to coordinate the settling and tell who needs to transfer what amount etc. Its mostly needed for groups that came together once in a while, hence settling needs to be done on group level.

Also on the same page if we can have a button that would calculate the minimum amount of transfers to settle everyone (cash flow minimization). I found an algorithm in below link, same can be used to do the calculation. To try, I downloaded json data from the app for one of our groups and used the same algorithm to reduce 15 transactions down to 5. 

https://www.geeksforgeeks.org/minimize-cash-flow-among-given-set-friends-borrowed-money/

krokosik:
Closes #158
Closes #58

- added personal balances on group page underneath the total balance
- added a balance list page for group
- added a simplify debts toggle in group options and a database field
- implemented a debt simplification algorithm based on [this article](https://www.geeksforgeeks.org/minimize-cash-flow-among-given-set-friends-borrowed-money/)
- created some unit tests and added Jest
- moved the expense list into a component shared with both the firend and group page
- added new shadcn aliases for compatibility with newer CLI versions
- moved some devDependencies from prod dependencies in `package.json`
- fixed some missing keys console errors due to incorrect usage of React fragments
- fixed a dev console error on missing Web Push key throwing an undefined query result error


krokosik:
I extended the test suite and noticed that the algorithm is not yet fully correct, any help in debugging will be appreciated.

PR #199, opened 2025-02-19
feat: add recalculate balances functionality and confirmation dialog

### LINKED ISSUE #198: Feature: recalculate balances button
We have been using Split Pro for a while, but it seems our balances are not properly calculated. In particular, all group balances are equal to zero despite having a multitude of expenses and I expect that issue #190 might be one of the reasons.

Since Split Pro is still under development and lacks a test suite that would give more confidence to its mechanism, I propose adding a recalculate balances button, available only to group admins. This would take all expenses and derive the group balances, without affecting anything outside from the `GroupBalance` table.

krokosik:
closes #198


krokosik:
I think this might not be the best solution and would like to discuss it in the issue, before moving forward

Issue #217, opened 2025-05-06
`web-push` not available in docker image


brunobeeee:
I want to setup web-push but I had to notice that the web-push binary isn't copied inside the docker image in `docker/Dockerfile`.
```Dockerfile
COPY --from=base  /app/prisma/schema.prisma ./prisma/schema.prisma
COPY --from=base  /app/prisma/migrations ./prisma/migrations
COPY --from=base  /app/node_modules/prisma ./node_modules/prisma
COPY --from=base  /app/node_modules/@prisma ./node_modules/@prisma
COPY --from=base  /app/node_modules/sharp ./node_modules/sharp
```

So you can't use web-push when running in docker.

Thanks for your work I really like SplitPro!

FriesischScott:
I'm running `split-pro` in `docker`  and push notifications work perfectly.

FriesischScott:
`split-pro` isn't using the `web-push` binary to send the notifications. It happens here: https://github.com/oss-apps/split-pro/blob/7f71b43c8f9fed8053ef76099542cde30b8ccb11/src/server/notification.ts#L12

FriesischScott:
Did you correctly set the environment variables:

- `WEB_PUSH_EMAIL}`,
- `WEB_PUSH_PUBLIC_KEY`,
-  `WEB_PUSH_PRIVATE_KEY`?

FriesischScott:
No I see what your issue is. You're supposed to generated the `web-push` key somewhere else before running the container not inside.

brunobeeee:
Oh yes you're right. I use docker only for splitpro and wanted to generate the keypair usind the container:

```bash
docker compose exec splitpro web-push generate-vapid-keys --json
```

This wouldn't work so I assumed that the `webpush` binary was missing inside the container and created PR #219.

Maybe it would be good to specify this in `README.md` or `.env.example` that it is not possible to generate the keypair inside the container?

brunobeeee:
**Idea**: This is a minmal command to generate the keypair using a docker container:
```
docker run --rm node:20-alpine3.19 npx web-push generate-vapid-keys --json
```

`node:20-alpine3.19` is the base layer of the split-pro image so you don't download the layers twice when running this command.

krokosik:
Added comment in commit d212c6529e21d7864a9026272a2e8092eb2eba77

PR #219, opened 2025-05-08
bugfix/issue-217-web-push-in-docker-image

brunobeeee:
In Issue #217 I noticed that `web-push` is not present in the docker image which prevented me from using push notifications. I guess due to shrinking the image size the multi-stage docker build installs all packages in the `base` stage and then copies only necessary files to the `release` stage.

See `Dockerfile`:
```Dockerfile
COPY --from=base  /app/prisma/schema.prisma ./prisma/schema.prisma
COPY --from=base  /app/prisma/migrations ./prisma/migrations
COPY --from=base  /app/node_modules/prisma ./node_modules/prisma
COPY --from=base  /app/node_modules/@prisma ./node_modules/@prisma
COPY --from=base  /app/node_modules/sharp ./node_modules/sharp
```

But this results in `web-push` not being available inside the `release` stage, so I added these lines that copy `web-push` and the `pnpm` dependencies of it.:
```Dockerfile
COPY --from=base  /app/node_modules/web-push ./node_modules/web-push
COPY --from=base  /app/node_modules/.pnpm/web-push@3.6.7 ./node_modules/.pnpm/web-push@3.6.7
```

For `web-push` to be runnable you also have to symlink the binary to `node_modules` similar to how the `prisma` binary was linked:
```Dockerfile
# Symlink the web-push binary
RUN ln -s /app/node_modules/web-push/src/cli.js ./node_modules/.bin/web-push
```

I would think that it now works but when running `npx web-push` inside the container, you get an error that the required dependency `minimist` is not present. So I copied it too:
```Dockerfile
# Required dependency for web-push
COPY --from=base  /app/node_modules/.pnpm/minimist@1.2.8 ./node_modules/.pnpm/minimist@1.2.8
```

I also added a short comment to `.env.example` on how to generate `web-push` keys when using docker.
I hope you like my solution, thank you for developing splitpro!

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -60,6 +60,7 @@ AUTHENTIK_ISSUER=
 
 # Push notification, Web Push: https://www.npmjs.com/package/web-push
 # generate web push keys using this command: web-push generate-vapid-keys --json
+# when using docker use this command: docker compose exec splitpro npx web-push generate-vapid-keys --json
 WEB_PUSH_PRIVATE_KEY=
 WEB_PUSH_PUBLIC_KEY=
 WEB_PUSH_EMAIL=
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -43,6 +43,14 @@ COPY --from=base  /app/node_modules/sharp ./node_modules/sharp
 RUN mkdir node_modules/.bin
 RUN ln -s /app/node_modules/prisma/build/index.js ./node_modules/.bin/prisma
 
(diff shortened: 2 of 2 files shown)


brunobeeee:
I assumed that `web-push` was missing inside the image bc I couldn't generate the keypair inside the container. But apparently you're supposed to generate the keypair somwhere else and the container runs w/o `web-push` binary, so this PR is obsolete.

PR #220, opened 2025-05-12
Add new group members to existing expenses

### LINKED ISSUE #216: New group members can't be added to existing expenses.
Currently, existing expenses can't be updated for new members when someone is added to the group. It would be great if the edit dialog included members that weren't present when the expense was originally created.

I don't think it should be too difficult to implement and would be happy to do it myself. I see two ways of doing this:

1. if a new member is added to a group all expenses in the database are updated to include them in the split without changing the existing distribution. 
2. When editing an expense additional members not present in the split are loaded so they can be  included in the updated expense

I prefer the first. I think it's quite clean.

@KMKoushik What do you think?

FriesischScott:
Currenttly new members cannot be added to existing expenses by modifying the expense. This issue is fixed by adding new group members to existing group expenses.

For safety, expenses are filtered to ensure we don't try to add a user to an expense twice. I don't really expect this to happen, but it's better to be safe than sorry.

Closes #216.

### DIFF
--- a/src/server/api/routers/group.ts
+++ b/src/server/api/routers/group.ts
@@ -274,12 +274,42 @@ export const groupRouter = createTRPCRouter({
   addMembers: groupProcedure
     .input(z.object({ userIds: z.array(z.number()) }))
     .mutation(async ({ input, ctx }) => {
-      console.log(input.userIds);
       const groupUsers = await ctx.db.groupUser.createMany({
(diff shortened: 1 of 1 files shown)


FriesischScott:
Since I now agree with @krokosik  that this is best done in the frontend I'm closing this.

Issue #234, opened 2025-05-28
Deleting or modifying group expenses updates balances incorrectly


FriesischScott:
There is an issues where group balances are incorrect after deleting an expense or changing the currency.

In my case, one group member's balance was off exactly his portion of a deleted expense. In another case we updated an expense from EUR to USD and the USD where still shown for some users.

I've verified that the data in `ExpenseParticipant` was correct in both cases.

Is there a specific reason for tracking expenses and balances separately? As we can see, keeping track of  one expense in four tables (`Expense`, `ExpenseParticipant`, `Balance` and `GroupBalance`) is difficult. All the info is in the `ExpenseParticipant` table and the balances could be efficiently computed in the database. 

For example, this query computes the balance for user 9 in group 3.

```sql
select sum("ExpenseParticipant".amount) from "Expense" join "ExpenseParticipant" on id = "expenseId" where "groupId"=3 and "userId"=9 and "deletedAt" is null;
```

With some additional `group by` magic all balances for an entire group can be calculate and I'm sure `prisma` can do the same.

If you're interested I'd be happy to give this a go and prepare a PR.

FriesischScott:
Sounds like a good plan. I'll prepare a proof of concept and we'll see where it goes from there.

krokosik:
If this works out I guess it will make #198 obsolete. My main concern is the performance, I wonder what would be the impact on a locally deployed Postgres database when we have say 30 members with a history of a 10k, 100k, 1M expenses.

krokosik:
@FriesischScott @guenzd I have created a PoC balance SQL query in branch `remove-balances`. I found Prisma too limiting and decided to go for TypedSQL queries, however I am still worried about performance and will try working on some stress tests. Let me know what you think about my approach and I'd be happy to hear your opinions.

FriesischScott:
> [@FriesischScott](https://github.com/FriesischScott) [@guenzd](https://github.com/guenzd) I have created a PoC balance SQL query in branch `remove-balances`. I found Prisma too limiting and decided to go for TypedSQL queries, however I am still worried about performance and will try working on some stress tests. Let me know what you think about my approach and I'd be happy to hear your opinions.

I''ve also opted for `typedSQL` it works great.  So far I've implemented group balances for the group overview and member balances for individual groups. You can find it here: https://github.com/FriesischScott/split-pro/tree/sql-balances

FriesischScott:
I'll open a draft PR with my changes tomorrow for better review.

[...]

Issue #207, opened 2025-03-12
How can I import my full spending history from Splitwise?


josefhelie:
Hi and thanks for your work! I’m in the process of onboarding to SplitPro and just exported a dataset from Splitwise to convert my Tricount data to that JSON format. However, I just realized that only balances were exported, not the individual transactions.

Is there a way to import the full transaction history? My girlfriend and I would love to keep all the history—and all the little jokes we wrote while spending our money! 😊

Update: I opened the database, but my SQL skills are way too rusty to write a tool for importing an expense dataset into PostgreSQL. Can someone help me with this?

josefhelie:
Well, I coded a python script to import data from a splitwise dataset, the example works with two users:


```
Date,Description,Catégorie,Cost,Devise,user1,user2
2025-03-11,Company,General,"7,2",EUR,"3,60","-3,60"
,,,,,"3,60","--3,60"
```

It looks like this, if you wanna use it to inspire yourself:

```
import csv  # Module to handle CSV file reading and writing
import random  # Module for generating random numbers and selections
import string  # Module providing string constants (e.g., lowercase letters)
import sys  # Module for accessing command-line arguments
from decimal import Decimal, InvalidOperation  # For precise decimal arithmetic and error handling during conversions
from datetime import datetime  # To obtain the current date and time

def generate_id():
    """
    Generates an identifier that matches the regex pattern:
    ^cm84[a-z]{5}000[a-z0-9]zmcd[a-z0-9]{8}$
    """
    part1 = "cm84"  # Fixed prefix
    part2 = ''.join(random.choices(string.ascii_lowercase, k=5))  # 5 random lowercase letters
    part3 = "000"  # Fixed sequence "000"
    part4 = random.choice(string.ascii_lowercase + string.digits)  # A single random alphanumeric character
    part5 = "zmcd"  # Fixed string "zmcd"
    part6 = ''.join(random.choices(string.ascii_lowercase + string.digits, k=8))  # 8 random alphanumeric characters
    return part1 + part2 + part3 + part4 + part5 + part6  # Concatenate all parts to form the final identifier

def escape_string(s):
    """
    Escapes single quotes in SQL strings by doubling them.
    This is used to prevent SQL injection or errors in SQL queries.
    """
    return s.replace("'", "''")

def main(csv_file, sql_file):
    # Open the input CSV file for reading and the output SQL file for writing (both with UTF-8 encoding)
[...]

PR #247, opened 2025-06-09
WIP: Compute balances in database

FriesischScott:
As discussed in #234 the goal of this PR is to drop the separate balance tables for users and groups and instead compute the balances from all expenses in the database.

I agree with @krokosik  that we need to be certain we can handle the required loads for large instances with lots of expenses but I think postgresql will be up for the task.

This is a list of the required balances:
1. [ ] Friend balances
2. [x] Overall group balances for one user
3. [x] All balances for a specific group

here I've implemented the second and third and I think @krokosik  has started working on the first.

Because `prisma` can't handle the necessary computations through it's javascript interface I've chosen to use `typedSQL` to write native SQL queries and still benifit from prisma's typing capabilities.

One minor downside is that `typedSQL` currently requires a live database connection to generate the types. However with `prisma` 7 the location of the generated client will move out of `node_modules` and we will have to set an output directory like [this](https://www.prisma.io/docs/orm/prisma-client/setup-and-configuration/generating-prisma-client#using-a-custom-output-path).

We could then commit the generated sql types in the `sql` subfolder of the prisma client and only generate the regular client during build. This should already be possible now (only in prisma 7 it's mandatory) and I'm working on setting it up.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -11,7 +11,7 @@ jobs:
       DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/postgres"
     services:
       postgres:
-        image: postgres
+        image: ossapps/postgres
         env:
           POSTGRES_PASSWORD: postgres
         options: >-


prisma/seed.ts
```diff
@@ -1,4 +1,4 @@
-import { PrismaClient } from '@prisma/client';
+import { PrismaClient } from '~/prisma/client';
```
krokosik: Let's minimize the changes in this PR by creating an alias for this import in `tsconfig.json`

prisma/schema.prisma
```diff
@@ -3,7 +3,8 @@
 
 generator client {
     provider        = "prisma-client-js"
-    previewFeatures = ["relationJoins"]
+    output = "../src/prisma/client"
```
krokosik: I would prefer that autogenerated files are not commited to VCS. Since SplitPro does not work without a databse, it does not hurt us to generate these at runtime, but I will double check

FriesischScott: The alternative is to have a database running in github actions and during docker build which is a lot more difficult than to include the files.

FriesischScott: Everything that can be generated without the database should of course not be committed.

krokosik: Not really, see the `start.sh` script

FriesischScott: The `start.sh` script runs the migrations at which point we can obviously expect a database to be running, but we also need the generated  files for `typedSQL` you get from `prisma generate --sql`  during build time.

I believe committing these files to the repository is the cleanes workaround for now. Here's a link to the ongoing discussion on the matter: https://github.com/prisma/prisma/issues/25124

guenzd:
I get this error when I try to run with existing data:
grafik

FriesischScott:
> I get this error when I try to run with existing data: grafik

Yes, I haven't finished fixing a few type issues. You should be able to run it with `pnpm dev` but `build` is not yet working.

prisma/seed.ts
FriesischScott: I didn't know that was a thing, but that sounds good, I'll look into it.

guenzd:
> > I get this error when I try to run with existing data: grafik
> 
> Yes, I haven't finished fixing a few type issues. You should be able to run it with `pnpm dev` but `build` is not yet working.

Then there are more runtime issues :/

grafik

prisma/seed.ts
FriesischScott: Good call, much cleaner. ✔️

prisma/schema.prisma
krokosik: Ah of course, you're right. Thanks for posting the issue I can see they also have a workaround that generates the files inside the Docker image? I would prefer to go with this approach, the time needed for the sleep should be pretty much reproducible on a Github CI runner

FriesischScott: I think that can be done. I'll make it happen once I fixed the remaining issues to get the build working.

FriesischScott: Stupid me... Of course we can run a postgres service container for the actions.

In the Docker build we can then rely on the workaround.

FriesischScott:
> > > I get this error when I try to run with existing data: grafik
> > 
> > 
> > Yes, I haven't finished fixing a few type issues. You should be able to run it with `pnpm dev` but `build` is not yet working.
> 
> Then there are more runtime issues :/
> grafik

Can you try again now? 

Turns out I wasn't up to date on the migrations and for some reason when the column is `bigint` the sum is returned as `numeric`. I'm now casting explicitly to `bigint`.

FriesischScott:
Not sure why the linting fails when it doesn't for me but otherwise it should now build without errors.

FriesischScott:
I've changed the `seed.ts` so you can use it check the performance yourself.

I think with roughly 80ms for 100k expenses we are well within the performance we require. I don't believe any group will reach that high.

FriesischScott:
We have lift off 🥳. The path alias was confusing the linter so I've moved the prisma client back into node_modules.

FriesischScott:
I've updated the seed file to create 1000 users and 101 groups with 10k expenses each. The largest group has 30 members all the others have 10. It should give us a better understanding of the performance.

It does take quite a while to create all the expenses but it's a temporary change anyway. If we merge we can just revert the file completely.

krokosik:
Fantastic work, thank you 🙌
I know I promised to take a swing at it this week, but unfortunately it's not likely to happen due to other obligations :/

FriesischScott:
> Fantastic work, thank you 🙌 I know I promised to take a swing at it this week, but unfortunately it's not likely to happen due to other obligations :/

No problem at all. Take your time. Better not to rush things :)

prisma/seed.ts
krokosik, line 1: The original `seed` script is useful for local dev work, while this one takes a long time to execute. Please move it to a separate file

prisma/sql/getAllBalancesForGroup.sql
krokosik, line 1: Can you apply some SQL formatting to these queries?

src/lib/simplify.test.ts
```diff
@@ -9,46 +8,45 @@ type MinimalEdge = {
...
 
-const sortByIds = (a: GroupBalance, b: GroupBalance) => {
-  if (a.userId === b.userId) {
-    return a.firendId - b.firendId;
+const sortByIds = (a: getAllBalancesForGroup.Result, b: getAllBalancesForGroup.Result) => {
+  if (a.paidBy === b.paidBy) {
+    return a.borrowedBy - b.borrowedBy;
   }
-  return a.userId - b.userId;
+  return a.paidBy - b.paidBy;
```
krokosik, line 15: I would prefer to keep this PR as minimal as possible. As such, please keep the property names identical and create a `type GroupBalance = getAllBalancesForGroup.Result` alias

src/server/api/routers/group.ts
```diff
@@ -56,43 +57,32 @@ export const groupRouter = createTRPCRouter({
...
+      const balancesForGroup: Record<string, bigint> = {};
+      groupBalances
+        .filter((b) => {
+          return b.id == group.id;
+        })
+        .forEach((b) => {
+          if (b.currency != null && b.balance != null) {
+            balancesForGroup[b.currency] = b.balance;
+          }
+        });
```
krokosik: While we remain on the dev branch, I propose the following sanity check:
utilize both the legacy and new method and compare their results. When a mismatch is detected, throw an Error. We will remove it before creating a release.

krokosik:
Oh and one more thing, this change would break the current splitwise import functionality, as it does not import expenses as of now. #207 needs to be implemented first and as such I don't expect to go ahead with this PR for 1.5 release

FriesischScott:
> Oh and one more thing, this change would break the current splitwise import functionality, as it does not import expenses as of now. #207 needs to be implemented first and as such I don't expect to go ahead with this PR for 1.5 release

No problem. More time to polish.

prisma/seed.ts
FriesischScott: My plan was to only use this as and intermediate script and revert to the original before we merge. If you think it's worth keeping around I'll move it to a separate script.

src/lib/simplify.test.ts
FriesischScott: Just to confirm, you want me to keep the existing `firendId` including the typo?

alexanderwassbjer:
This is really good work! Calculating it based on the expenses rather then the balance table is amazing! 👏

FriesischScott:
Thanks for the ping. I'm of course happy to finish this. 

I'll start by merging in all the changes and fixing the conflicts. Shouldn't take more than a day or two until I have something up to date and working.

krokosik:
Okay nice, since I would like to include my changes as well in this PR and have some thoughts I would like to share, would you like to connect on Discord to chat more conveniently?

FriesischScott:
I've fixed the conflicts and properly formatted the SQL files. So far everything still seems to work as intended. The check currently fails because Github Actions doesn't allow to change the command for the postgres container to load the pg_cron extension.

The easiest fix would probably be if you could update the custom postgres image to contain the `shared_preload_libraries`.

FriesischScott:
> Okay nice, since I would like to include my changes as well in this PR and have some thoughts I would like to share, would you like to connect on Discord to chat more conveniently?

Yeah, I think that would help. I'll email you my username.

alexanderwassbjer:
> I've fixed the conflicts and properly formatted the SQL files. So far everything still seems to work as intended. The check currently fails because Github Actions doesn't allow to change the command for the postgres container to load the pg_cron extension.
> 
> 
> 
> The easiest fix would probably be if you could update the custom postgres image to contain the `shared_preload_libraries`.

Good work! Looking forward to this change.

src/lib/simplify.test.ts
krokosik: While I agree the naming even without the typo is quite unfortunate, I would prefer to make such a critical PR as small and easy to review as possible. We can think about another PR afterwards.

krokosik:
@FriesischScott I've sent you a friend request. I am now working on an extensive seeding script with `fakerjs` and some sociale network graph generation algorithms. It would be nice to have a deterministic dev db and we could establish some data consistency check on it and more importantly compare the existing balance tables with the query calculated ones.

FriesischScott:
> @FriesischScott I've sent you a friend request. I am now working on an extensive seeding script with `fakerjs` and some sociale network graph generation algorithms. It would be nice to have a deterministic dev db and we could establish some data consistency check on it and more importantly compare the existing balance tables with the query calculated ones.

Good plan!

krokosik:
The build error seems to stem from the fact you don't provide startup commands for cron. See the repo compose files for reference.

krokosik:
@FriesischScott The seed script has been merged :)
It should serve as a replacement over your `largeSeed` file and is more nuanced with edited and deleted expenses.
You can tweak the number of users and groups to get more expenses and I recommend checking out the `CONTRIBUTING.md`. Experiment with the params by calling `pnpm tsx src/dummies` to get a `SEED_STATISTICS.md` report until you have some desirable data.

The next step would be to implement checks. Currently we still create balances, which also prevents us from inserting expenses in parallel as all the transactions create deadlocks. I would propose writing some consistency checks, that would compare as many cases of query calculated balances to the table ones as possible, preferably all of them.

Once that is done, we can plan next steps, as the situation with direct user-user balances requires us to think through the transition from current Total balances to user-user ones.

FriesischScott:
Closing as obsolete, see #479.

PR #263, opened 2025-07-01
Add BigInt JSON replacer for account data export

marcosinigaglia:
You can't serialize BigiInt. So I added a bigIntReplacer function to serialize BigInt values as strings in JSON output.


krokosik:
Thank you for the PR, please fix the formatting and we can merge this :)

PR #295, opened 2025-08-06
Add German locale

soeffi:
Added German localization files


krokosik:
Awesome work, thank you!

PR #296, opened 2025-08-07
Avoid redirect loop

marcosinigaglia:
If `DEFAULT_HOMEPAGE` is set to `/home` and you are already authenticated, you cannot exit the home -> login -> home loop.


krokosik:
good catch, thank you. Could you also make another PR for the release-1.4 branch?

Issue #294, opened 2025-08-05
i18next seems to load only one namespace


krokosik:
**Describe the bug**
When trying to use both `common` and some other namespace translations, we get missing strings.

**To Reproduce**
1. Go to details of some expense
2. See that everything except for the _you_ string is correctly translated.

**Screenshots**
<img width="410" height="585" alt="Image" src="https://github.com/user-attachments/assets/e396a926-d289-4fe0-a939-bc9c138596cc" />

**App version**
main branch

**Additional context**
It looks like the issue is caused by `ui.you` belonging to `common`, while the rest of expenseDetails utilizes its own namespace. Curiously, when the order of namespaces changes in the `useTranslationWithUtils` hook, the other namespace stops working.

@marcosinigaglia since you implemented the feature, could you take a look at it?

marcosinigaglia:
I'll take a look.

PR #297, opened 2025-08-07
Fix/displayname

marcosinigaglia:
Fix the issue https://github.com/oss-apps/split-pro/issues/294


krokosik:
Just for my understanding, why is it needed for two namespaces? I assume that the first one in the array is the default and if we need the other one, then we need to explicitly specify it?

marcosinigaglia:
I don't know why, but it doesn't work unless you specify to use the 'common' namespace, even if it has already been loaded. It's the only place where it's needed.

PR #305, opened 2025-08-11
Split remaining equally

### LINKED ISSUE #287: Add split up remaining button when using exact unequal spljt
**Is your feature request related to a problem? Please describe.**
When ordering food the delivery fees are included in the total price but not when adding up the individual items

Currently it displays the remainder but it should also have a button to equally split it up (as far as far possible)

D3SOX:
Resolves #287

> Disclaimer:
This is fully AI generated but I tested it and it seemed to work fine.
(first time trying GPT-5 in Cursor)

Let me know if there are any issues with the code.


public/locales/de/expense_details.json
```diff
@@ -21,6 +21,8 @@
         "split_unequally": "ungleichmäßig teilen",
         "save": "Speichern",
         "all": "Alle",
+        "confirm": "Bestätigen",
```
krokosik, line 24: Something I missed in the original i18 PR, stuff like this should go to common, but it's sth for another PR.

D3SOX: Yeah sorry I mixed in a bit of stuff :D
Actually tried to use ui.cancel everywhere because it's duplicated.

src/store/addStore.ts
```diff
@@ -99,13 +104,10 @@ export const useAddExpenseStore = create<AddExpenseState>()((set) => ({
...
-          ...state.splitShares,
-          [userId]: {
-            ...state.splitShares[userId],
-            [splitType]: share,
-          },
-        } as SplitShares;
+        const splitShares: SplitShares = { ...state.splitShares };
+        const userShares = splitShares[userId] ?? initSplitShares();
+        userShares[splitType] = share;
+        splitShares[userId] = userShares;
```
krokosik, line 110: What is the purpose of this change?

src/store/addStore.ts
```diff
@@ -343,9 +411,14 @@ export function calculateParticipantSplit(
...
-  // @ts-expect-error TS enums/string coersion *eyeroll*
-  Object.fromEntries(Object.values(SplitType).map((type) => [type, undefined]));
+export const initSplitShares = (): Record<SplitType, bigint | undefined> => ({
+  [SplitType.EQUAL]: undefined,
+  [SplitType.PERCENTAGE]: undefined,
+  [SplitType.SHARE]: undefined,
+  [SplitType.EXACT]: undefined,
+  [SplitType.ADJUSTMENT]: undefined,
+  [SplitType.SETTLEMENT]: undefined,
+});
```
krokosik, line 420: What is the purpose of this change?

src/components/AddExpense/SplitTypeSection.tsx
```diff
@@ -276,13 +333,134 @@ const SplitSection: React.FC<SplitSectionProps> = ({
...
+              {t('ui.add_expense_details.split_type_section.types.exact.split_remaining_equally')}
+            </span>
+          </button>
+        ) : null}
       </div>
+
+      <Dialog open={showConfirmModal} onOpenChange={setShowConfirmModal}>
+        <DialogContent className="sm:max-w-md">
+          <DialogHeader>
+            <DialogTitle>
```
krokosik, line 381: I have two big issues with the modal. First a modal within a modal is a bit much and should only be used when really necessary. Second of all, it looks nothing like any component in the rest of the app (even though it does not look bad). This is espcially bad from UI/UX standpoint, when we already have a component for splitting an amount equally and it is completely different.

I would propose the following to solve both problems:
we already have a participant select screen in the EQUAL split section. We should reuse this styling AND do not display it in a modal. By the latter I mean staying in the same modal, but instead of the exact split inputs, show the boolean toggles, like so:

1. User navigates to EXACT split, sees the typical view, with numerical inputs AND a split remainder button
2. Upon clicking, the button disappears. The user is still in the EXACT split tab, but now the choice is boolean:
  - the remaining value to be distributed text is displayed
  - the buttons at the bottom are changed
  - maybe by default, users with non zero amounts are selected for the split
3. Confirming adds the calculated amounts to the split amounts, canceling does nothing and takes us back.
4. It should be possible to navigate to other split sections. The split remainder equally view should be disabled then.

src/pages/balances/[friendId]/expenses/[expenseId].tsx
```diff
@@ -69,7 +69,7 @@ ExpensesPage.auth = true;
 
 export const getServerSideProps: GetServerSideProps = async (context) => ({
   props: {
-    storagePublicUrl: env.R2_PUBLIC_URL,
+    ...(env.R2_PUBLIC_URL ? { storagePublicUrl: env.R2_PUBLIC_URL } : {}),
```
Comment (krokosik) [inline src/pages/balances/[friendId]/expenses/[expenseId].tsx:72]: What is the purpose of this change?

src/utils/numbers.ts
```diff
@@ -91,5 +91,50 @@ export const BigMath = {
   },
 };
 
+/**
```
krokosik, line 94: Will it be reused anywhere? I don't think so, it should stay in addStore like other split related calculations.

krokosik (changes requested):

public/locales/de/expense_details.json
krokosik: No worries, that's something that we had in the code before :)

src/pages/balances/[friendId]/expenses/[expenseId].tsx
Comment (D3SOX) [reply src/pages/balances/[friendId]/expenses/[expenseId].tsx:72]: Will revert might only be some fluke that I thought I need it locally

src/store/addStore.ts
D3SOX: Tried to improve the types but I agree the former is more readable, will revert

src/store/addStore.ts
D3SOX: Tried to improve the types but I agree the former is more readable, will revert

krokosik:
@D3SOX are you going to continue work on this? If not I am going to close this PR as stale.

krokosik:
Closing as abandoned and far from mergeable

PR #306, opened 2025-08-12
🇸🇪 Add swedish language

alexanderwassbjer:


krokosik:
Toppen! Tack så mycket 🇸🇪 
once the CI checks are passing, we can merge this :)

alexanderwassbjer:
@krokosik :) Now I fixed so the CI checks are passing

krokosik:
@alexanderwassbjer it somehowe slipped my mind that Sweden's ISO code is SE. How did that SV happen? 😅

alexanderwassbjer:
> @alexanderwassbjer it somehowe slipped my mind that Sweden's ISO code is SE. How did that SV happen? 😅

😁 The normal ISO code for Sweden is SE indeed but the swedish locale regarding to i18n standard should be sv. And if you have multiple languages in swedish it can be sv-SE. But because we only have one language currently it's better with sv.
It's really confusing but actually sv should be correct 😅

Quote: "The ISO 639-1 language code for Swedish is sv. For example, in i18next, you would use sv to specify the Swedish language"

krokosik:
I see, when I ran it with i18ally, I got an unknown to me flag. What would happen if we change it to SE?

PR #327, opened 2025-08-25
Add i18n for European Portuguese

c42759:
Adds Portuguese translation files for account, categories, common, currencies, expense details, groups details, home, and signin pages.
This provides a fully translated user interface for Portuguese-speaking users.


public/locales/pt/common.json
```diff
@@ -0,0 +1,89 @@
...
+    "balances": "Balancetes",
+    "groups": "Grupos",
+    "add_expense": "Adicionar Despesa",
+    "add": "Adicionar",
+    "activity": "Atividade",
+    "account": "Conta"
+  },
+  "ui": {
+    "actors": {
+      "you": "Tu",
```
krokosik, line 18: Missing keys for dative and accusative cases.

public/locales/pt/expense_details.json
```diff
@@ -0,0 +1,73 @@
...
+            "remaining": "Restante"
+          },
+          "exact": {
+            "title": "Exato",
+            "remaining": "Restante"
+          },
+          "share": {
+            "title": "Share",
+            "total_shares": "Total shares",
+            "shares": "Share(s)"
```
krokosik: Is that portugese?

public/locales/pt/groups_details.json
```diff
@@ -0,0 +1,76 @@
...
+      },
+      "invite_link": "Partilhar link de convite",
+      "copied": "Copiado"
+    },
+    "group_statistics": {
+      "title": "Estatísticas do grupo",
+      "total_expenses": "Total de despesas",
+      "first_expense": "Primeira despesa"
+    },
+    "group_info": {
```
krokosik, line 23: Missing new archive keys

public/locales/pt/currencies.json
```diff
@@ -0,0 +1,559 @@
+{
+  "title": "Selecionar moeda",
+  "placeholder": "Selecionar moeda",
+  "no_currency_found": "Moeda não encontrada.",
+  "currency_list": {
```
krokosik, line 5: This should also be in Portugese

krokosik (changes requested):
There are some missing keys. Also include the new locale in the `i18n.ts` file.

public/locales/pt/groups_details.json
c42759: Fixed

public/locales/pt/common.json
c42759: Fixed

public/locales/pt/expense_details.json
c42759: Sorry, this was a bit tricky.
In Portuguese without context this would be like sharing a Post from Social media, and I had to look where this was the use case, and is translated to a totally different word.

c42759 (commented):
Just place the fixes based on your comments @krokosik

public/locales/pt/currencies.json
c42759: Finished  😁

public/locales/pt/expense_details.json
krokosik: In general I would recommend to go over the app on your machine and check that most stuff looks okay. For example, you might notice that Portugese will be missing from the language selector ;)

https://github.com/oss-apps/split-pro/blob/d30c81c980679d9a8bc10befadf06b2c79b0da7a/src/utils/i18n/client.ts#L9

c42759:
Took me a while to get the local environment running, but I was able to check it 😊

PR #334, opened 2025-08-27
Use correct case (dativus) in translations

soeffi:
Identified two locations which should use "dativus" case for "you"


krokosik:
Ah thank you, it made me realize I put in the cases the other way around 😅

Issue #340, opened 2025-08-29
Bug: ui. at the start of some texts


Silther:
**Describe the bug**

<img width="1664" height="952" alt="Image" src="https://github.com/user-attachments/assets/17bbc322-345a-4c04-a91c-3af29b43a916" />

<img width="299" height="160" alt="Image" src="https://github.com/user-attachments/assets/15ae3335-48ad-439e-bdf5-cf7b136a67f4" />

**To Reproduce**
1. update to the latest beta
2. go to account

**Expected behavior**
most like a bug with the translations

**Screenshots**

**App version**
1.5.0 beta 4 

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Vivaldi (Chromium)

**Additional context**
sometimes it disappears and is displayed correctly on changing the page or refreshing.

krokosik:
Can you share some more details about your setup? Are you reaching it on your machine? A VM in the cloud? Behind a reverse proxy or localhost? I cannot reproduce it on my end.
Check if there are any errors in the console or failed network requests.

krokosik:
Can you check if the `fix-340` tag works?

Silther:
> Can you share some more details about your setup? Are you reaching it on your machine? A VM in the cloud? Behind a reverse proxy or localhost? I cannot reproduce it on my end.

I running it on docker compose on local home lab server and using cloudflared instead of my own reverse proxy.

Silther:
I believe it most likely was a cashing issue when translation was added, I cleared the browser data everywhere and now it seems to be gone.

krokosik:
Okay, but if the issue persists, you can try the build I sent.

Silther:
I already did, and had the same issue until I cleared the side data.

(But the add members button works again, so thanks :)

Silther:
On iOS in my safari pwa it worked after the 340 fix even without deleting the site data.
At first it still showed ui. but if I visited those pages again it was rendered correctly.

Silther:
I found some strings which still show the issue or are not yet translated:

Settle up:
<img width="427" height="430" alt="Image" src="https://github.com/user-attachments/assets/cbd6d6d7-57f4-4d13-a6fc-1ac37ce15c0f" />

krokosik:
Yes, it's a bug. Will fix it soon

PR #349, opened 2025-08-29
Include localePath in full config

krokosik:
Related to #340 

Can't reproduce so don't know if it's the solution, but I found this thread: https://stackoverflow.com/questions/69656025/next-i18next-is-not-working-with-serversideprops-in-dynamic-pages-after-deployin
and noticed we aren't actually using the localePath

### DIFF
--- a/next-i18next.config.js
+++ b/next-i18next.config.js
@@ -1,11 +1,13 @@
-/** @type {import('next-i18next').UserConfig} */
+import { resolve } from 'path';
+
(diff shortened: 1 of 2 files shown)


krokosik:
Actually, nextjs config doesn't have that option, it's just for i18n. I noticed I omitted the `trailingSlash: true` option from Internationalization docs: https://nextjs.org/docs/pages/guides/internationalization#prefixing-the-default-locale

PR #352, opened 2025-08-31
Fix hook error and add PWA install translations and localize InstallApp

marcosinigaglia:
There was an error in the `InstallApp` component in the PWA version. I fixed it and added translations.


public/locales/de/pwa.json
```diff
@@ -0,0 +1,19 @@
...
+      "title": "App herunterladen",
+      "left_action": "Schließen",
+      "description": "Du kannst SplitPro als PWA auf deinem Startbildschirm installieren",
+      "ios": {
+        "text": "Wenn du iOS verwendest, schau dir dieses",
+        "link_text": "Video"
+      },
+      "android": {
+        "text": "Wenn du Android verwendest, schau dir dieses",
+        "link_text": "Video"
```
krokosik: No need for a separate namespace if it is that small, in my opinion. Not only are you duplicating the `or` key, but the required translations are already in `account`

src/components/InstallApp.tsx
```diff
@@ -20,6 +22,16 @@ const InstallApp: React.FC = () => {
...
 
+  const DownloadButton = useMemo(
+    () => (
+      <Button className="w-[250px]">
+        <Download className="mr-2 h-5 w-5 text-black" />
+        {t('ui.download_button')}
+      </Button>
+    ),
+    [t],
+  );
```
krokosik: Let's refactor this and unite with the button from `account` page. If the trigger is passed as the `children` prop, we can simply reuse this component for both cases.

src/components/InstallApp.tsx
```diff
@@ -1,10 +1,12 @@
 import { Download } from 'lucide-react';
 import { useEffect, useMemo, useState } from 'react';
+import { useTranslation } from 'next-i18next';
 
 import { Button } from './ui/button';
 import { AppDrawer } from './ui/drawer';
 
 const InstallApp: React.FC = () => {
+  const { t } = useTranslation('pwa');
   const [isStandalone, setIsStandalone] = useState(false);
```
krokosik: I think this is a perfect use case for us to make an `useIsPwa` hook

marcosinigaglia:
Hi @krokosik , thanks for the reviews. I think is fine now.

src/components/Account/DownloadAppDrawer.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { AppDrawer } from '../ui/drawer';
+
+interface DownloadAppDrawerProps {
+  trigger?: ReactNode;
+  className?: string;
+}
+
+export const DownloadAppDrawer: React.FC<DownloadAppDrawerProps> = ({ trigger, className }) => {
+  const { t } = useTranslation('account_page');
+
```
krokosik: Let's check for PWA here and return null if so.

src/components/InstallApp.tsx
```diff
@@ -1,84 +1,19 @@
...
-            >
-              Video
-            </a>
-          </p>
-        </div>
-      </AppDrawer>
-
-      <p>or</p>
+      <DownloadAppDrawer />
+      <p>{t('ui.or', { ns: 'common' })}</p>
```
krokosik: Please define it on the balances page as a trigger component. No need for the almost empty InstallApp.

src/components/Account/DownloadAppDrawer.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Button } from '../ui/button';
+import { AppDrawer } from '../ui/drawer';
+
+interface DownloadAppDrawerProps {
+  trigger?: ReactNode;
+  className?: string;
+}
+
+export const DownloadAppDrawer: React.FC<DownloadAppDrawerProps> = ({ trigger, className }) => {
```
krokosik: It would be nicer (and more JSX/perf friendly) to use the children prop.

marcosinigaglia:
@krokosik done

src/pages/balances.tsx
```diff
@@ -117,7 +119,8 @@ const BalancePage: NextPageWithUser = () => {
 
             {!balanceQuery.isPending && !balanceQuery.data?.balances.length ? (
               <div className="mt-[40vh] flex -translate-y-[130%] flex-col items-center justify-center gap-6">
-                <InstallApp />
+                <DownloadAppDrawer />
+                {!isPwa && <p>{t('ui.or', { ns: 'common' })}</p>}
```
krokosik, line 128: This check is already inside the component. Simply add: <Button /><or /> as children.

src/components/Account/DownloadAppDrawer.tsx
```diff
@@ -4,14 +4,16 @@ import { useTranslation } from 'next-i18next';
 
 import { Button } from '../ui/button';
 import { AppDrawer } from '../ui/drawer';
+import useIsPwa from '~/hooks/useIsPwa';
 
 interface DownloadAppDrawerProps {
-  trigger?: ReactNode;
+  children?: ReactNode;
```
krokosik: I believe the children and className props are already inside React.FC?

marcosinigaglia: I didn't understand the request.

src/pages/balances.tsx
marcosinigaglia: The alignment breaks if they are not separate elements.

src/components/Account/DownloadAppDrawer.tsx
krokosik: Ok my bad, it's not included in FC

PR #364, opened 2025-09-05
Feat 79 currency rate transactions

### LINKED ISSUE #79: Feature Request: Allow currency exchange transactions
This would be useful on trips which include multiple countries.

Expenses and balances can be difficult to track when there are multiple currencies involved. For example on a trip to Japan I may be owed 100 USD for booking some accommodations from home and owing 10.000 JPY for my expenses covered by friends abroad. It's difficult to tell whether my balance is greater than zero or not. A relatively simple solution for this would be providing a view which dynamically converts all balances to a specified base currency on the fly, but that's not very robust due to the changing nature of rates.

Instead, I recommend implementing manual settlement transactions which convert one currency to another. The currency rate should be provided by the user (and maybe suggested by the application). 

In the example above once we leave Japan, we could convert the remaining balance of JPY to USD, so `+100 USD, -10.000 JPY` would become `+32 USD, 0 JPY` with a conversion rate of 146 JPY/USD (the settlement transaction would be `-68 USD, +10.000 JPY`)

krokosik:
Adding currency conversion transactions. In terms of APIs, we rely on the frankfurter app which is completely free and covers around 30 currencies. As a backup, one can provide an Open Exchange Rates App ID to cover transactions involving USD. Changing the base currency is sadly outside the scope of the free plan.

The API limits for OXR are mitigated a bit, as we fetch ALL the rates for a given rate in one call and then cache that day in the DB.

The first option covers all my needs, but if anyone is willing to add more providers, you are more than welcome to do so.

The UI still needs some polishing, which is why it is still a draft and I need to add `pg-cron` for cache clearing.

Locales also need to be added. A helping hand is more than welcome as I will be unavailable for a couple days.

Closes #79


krokosik:
https://github.com/user-attachments/assets/d163ece5-11d2-4882-bc5f-ce9f92f79712

autofix-troubleshooter:
Hi! I'm the <a href="https://autofix.ci">autofix logo<strong>autofix.ci</strong></a> troubleshooter bot.

It looks like you correctly set up a CI job that uses the autofix.ci GitHub Action, but the [**autofix.ci GitHub App**](https://github.com/apps/autofix-ci/) has not been installed for this repository. This means that autofix.ci unfortunately does not have the permissions to fix this pull request. If you are the repository owner, please install the app and then restart the CI workflow! 😃

PR #413, opened 2025-10-03
Allow >3-digit numbers in Exact section

masterpi007:


krokosik:
I'm sorry but the fix only scratched the surface of what was wrong with our currency input handling. I believe the new framework will be much more robust and I'm closing this PR in favor of #437

Issue #369, opened 2025-09-13
receipt missing


chihminwen:
Describe the bug
When an expense entry has a receipt photo attached, updating the entry causes the receipt to be deleted, leaving only the text record.

To Reproduce
	1.	Add a new expense entry and attach a receipt photo.
	2.	Save the entry and confirm the receipt is visible.
	3.	Edit the same entry (e.g., change the amount or note) and update it.
	4.	After updating, the previously attached receipt photo is missing.

Expected behavior
The receipt photo should remain attached after updating the entry, instead of being removed.

Screenshots

App version
v1.5.3

Desktop (please complete the following information):
	•	OS: macOS 15.6.1
	•	Browser Safari
	•	Version 18.6

Smartphone (please complete the following information):
	•	Device: iPhone 16
	•	OS: iOS 18.6.2
	•	Browser Safari
	•	Version 

Additional context
This may be caused by the update process overwriting the record without preserving the existing receipt field.

Spaceboy-ren:
hey! I would like to work on this issue and submit a PR. I am new to open source and looking forword to contribute:)

krokosik:
@Spaceboy-ren Nice to hear :)
I recommend starting with the README.md and CONTRIBUTING.md docs and feel free to reach out with any questions.

PR #417, opened 2025-10-04
Fixed the missing receipt issue after editing expense 

Spaceboy-ren:
This PR fixes the issue mentioned in #369.

Changes made:
- Added a query to fetch existing expense details when editing
- Used a useEffect hook to prefill all fields from expenseData
- Restored the fileKey so the uploaded receipt photo doesn’t disappear after editing
- Updated imports to include useEffect

Here is a video demonstrating the updated code:
https://github.com/user-attachments/assets/c2741ac6-fa4f-43ad-8275-1c08dd8f0f39

### DIFF
--- a/src/components/AddExpense/AddExpensePage.tsx
+++ b/src/components/AddExpense/AddExpensePage.tsx
@@ -44,6 +44,14 @@ export const AddOrEditExpensePage: React.FC<{
   const paidBy = useAddExpenseStore((s) => s.paidBy);
   const splitType = useAddExpenseStore((s) => s.splitType);
   const fileKey = useAddExpenseStore((s) => s.fileKey);
+  const { data: expenseData } = api.expense.getExpenseDetails.useQuery(
+    {
+      expenseId: expenseId as string,
+    },
+    {
+      enabled: !!expenseId,
+    },
+  );
 
   const {
     setCurrency,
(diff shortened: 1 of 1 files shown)


src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -54,11 +62,36 @@ export const AddOrEditExpensePage: React.FC<{
     resetState,
     setSplitScreenOpen,
     setExpenseDate,
+    setFileKey,
   } = useAddExpenseStore((s) => s.actions);
 
   const addExpenseMutation = api.expense.addOrEditExpense.useMutation();
   const updateProfile = api.user.updateUserDetail.useMutation();
 
+  React.useEffect(() => {
```
krokosik: Good call with what was missing! However, you may notice that there was no form filling logic in this component and only the image was missing. In fact, in you solution, it is still missing when editing the expense.

The place where we set the details and the fileKey is missing is in the `add.tsx` page :)

So what needs to be done is:
- add the setFileKey call to `add.tsx`
- change the condition in `UploadFile` component to propely show that the image is already added

Spaceboy-ren:
Ahh I see, thanks for your help!, now this is how it looks while adding an entry and while editing one.
Screenshot 2025-10-07 005801
Screenshot 2025-10-07 005905

Spaceboy-ren:
I messed up my wording in the previous response sorry, 

Adding expense:
image

Editing expense:
image



Is this what you meant by?
> Much better, but the photo icon in your second screenshot is still not in primary color when editing the expense attached.

krokosik:
Yes, looks good! There are also CI checks failing, once that is fixed we are good to merge :)

src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -66,6 +67,30 @@ export const AddOrEditExpensePage: React.FC<{
...
+    expenseData,
+    setDescription,
+    setAmount,
+    setAmountStr,
+    setCurrency,
+    setCategory,
+    setExpenseDate,
+    setFileKey,
+  ]);
+
```
krokosik: Nevermind, looks like I missed this. This change should be reverted, since it duplicates logic in `add`

Spaceboy-ren: So what do I exactly do to fix this issue? :)

krokosik: Remove your changes from `AddExpensePage`

src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -391,4 +391,4 @@ const SponsorUs = () => {
       </Link>
     </div>
   );
-};
+};
```
krokosik: It seems you did not apply autoformatting, which is also why CI keeps failing.

Issue #402, opened 2025-09-26
Left over pennies reshuffling


Silther:
**Describe the bug**
If you edit an expense the left over pennies get reshuffled between the participants.
This seems minor, but if multiple people already settled up, their previous settle up could be wrong by 1ct ([The leftover pennie](https://github.com/oss-apps/split-pro#how-are-leftover-pennies-handled)) which is an inconvenience because they now have to change the previous settled up sum by that amount. and if something gets edited again it could repeat ...

**To Reproduce**
1. Go to groups
2. Click on an expanse that couldn't be perfectly split and is already settled up
3. edit something
4. go to balances
5. see that some participants are no longer settled up

**Expected behavior**
un reshuffle the left over pennie if you create an expense edit the sum or modify the split settings.

**Screenshots**

<img width="892" height="165" alt="Image" src="https://github.com/user-attachments/assets/4450a0e3-ace4-459f-92f3-d6bd5a5b0324" />

**App version**
v1.5.3

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Vivaldi
 - Version 8.5

**Additional context**
In my case I changed the category of the item, but changing the name would lead to the same thing.

richardsokol:
Hi! I’d love to take on this issue and submit a PR. I’ve already gone through the code and implemented the fix but I’m still testing everything to make sure it works smoothly.

krokosik:
Of course :)
No need to ask for permission, just submit it

krokosik:
Any thoughts on making a deterministic shuffling method? This would mean that given a certain split configuration, always the same users would get pennies. I'm afraid that some may consider it unfair, but maybe it's worth it considering the issue you are metnion

Silther:
> Any thoughts on making a deterministic shuffling method? This would mean that given a certain split configuration, always the same users would get pennies.

Like using the integer that stores the money amount and than modulo the number of participants?
You could than iterate over an sorted array of the participants.


split-pro/tree/main/src/store/addStore.ts > calculateParticipantSplit
```typescript

if (canSplitScreenClosed) {
  let penniesLeft = updatedParticipants.reduce((acc, p) => acc + (p.amount ?? 0n), 0n);
  const participantsToPick = updatedParticipants.filter((p) => p.amount).sort();

  if (0 < participantsToPick.length) {
    let i : number = Number(amount % BigInt(participantsToPick.length));
      
    while (0n !== penniesLeft) {
[...]

PR #428, opened 2025-10-14
Fixed the left-over penny issue after editing expense #402

richardsokol:
# Change Left-Over Penny Distribution to be Deterministic

#402 Changes Made

### 1. Modified Penny Distribution Logic
- Replaced random penny distribution with a deterministic hash-based approach
- Introduced a hash function that takes into account:
  - Transaction ID
  - Last 6 digits of the amount
  - Participant ID
- The hash function ensures consistent but varied distribution based on these inputs
- For the same transaction (same ID and amount), the penny distribution will always be identical
- For different transactions, the distribution varies based on the transaction ID

### 2. Added Test Coverage
Added three new test cases to verify the behavior:

1. `should distribute leftover penny to different participants for different transactions`
   - Tests that different transactions can result in different penny distributions
   - Verifies that we get different recipients across multiple transactions

2. `should maintain consistent penny distribution for same transaction with different descriptions`
   - Verifies that modifying transaction metadata (title, emoji, category) doesn't affect the distribution
   - Ensures the same transaction ID always results in the same distribution

3. `should handle different transaction IDs with same amount`
   - Tests that different transaction IDs with the same amount can result in different distributions
   - Verifies the deterministic nature of the distribution while ensuring variability

## Purpose
This change improves the predictability and consistency of expense splitting by ensuring that:
- The same transaction will always split the same way, even after edits
- Different transactions distribute pennies differently
- The distribution is fair and based on transaction properties rather than random chance

## Testing
The new tests verify both the consistency within a transaction and variability across different transactions, ensuring the behavior is both deterministic and well-distributed.

### DIFF
--- a/src/store/addStore.test.ts
+++ b/src/store/addStore.test.ts
@@ -128,7 +128,25 @@ describe('calculateParticipantSplit', () => {
       expect(totalOwed).toBe(0n); // Total should always balance      // Payer should get the remainder
       expect(result.participants[0]?.amount).toBeGreaterThanOrEqual(6667n);
     });
+
+    it('should handle single participant as payer', () => {
+      const participants = createParticipants([user1]);
+      const splitShares = createSplitShares(participants, SplitType.EQUAL, [1n]);
+
+      const result = calculateParticipantSplit(
+        10000n,
+        participants,
+        SplitType.EQUAL,
+        splitShares,
+        user1,
+      );
+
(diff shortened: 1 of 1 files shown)


krokosik:
I am closing this as I don't have the capacity to properly review another verbose and incorrect AI generated PR. Your approach to make it deterministic might be a good idea, but the problem is transactionIds do not exist on most expenses.

PR #466, opened 2025-11-04
Leftover penny issue FIX

### LINKED ISSUE #402: Left over pennies reshuffling

richardsokol:
This PR closes #402 where settled balances would break after minor edits

Changes:
- Modified setParticipants() to detect when participants already have amounts from the database and skip recalculation
- Added tests to verify different penny distributions are preserved across edits 

Example:
Before: Split $0.04 three ways → edit description → pennies reshuffled
Now: Split $0.04 three ways → edit description → pennies stay the same

(Btw, Sorry for my previous incorrectly written PR with this issue)


krokosik:
After an initial glance, I have 3 major comments. Please take the time to set up the necessary tooling: prettier+oxc as clearly the formatting is broken. Second of all, the leftover pennies can also occur with percentage and share types which is not covered by this PR. Finally, I created a PR template for a reason, please use it.

krokosik:
Closed due to inactivity and not adhering to contributing rules

PR #471, opened 2025-11-05
Add locale-aware date formatting for UI dates

### LINKED ISSUE #470: Dates are not localized
**Describe the bug**
All dates currently show in English, regardless of the language chosen by the user

**To Reproduce**
1. Set app language to anything other than English
2. Open the expenses list or an expense details page

**Expected behavior**
Dates should be formatted according to the locale, with month abbreviations in the chosen language

**Screenshots**
App set to pt-BR but dates are all in English:
<table><tr><td>
Image
</td><td>
Image
</td></tr></table>

**App version**
main branch

rodrigost23:
# Description
Adds locale-aware dates, fixes #470.

# Demo
<table><tr><td>
image
</td><td>
image
</td></tr></table>

# Checklist

### DIFF
--- a/src/utils/strings.ts
+++ b/src/utils/strings.ts
@@ -1,9 +1,8 @@
 import { SplitType, type User } from '@prisma/client';
-import { format, isToday } from 'date-fns';
+import { isToday } from 'date-fns';
 import { type TFunction } from 'next-i18next';
 import { type CurrencyCode } from '~/lib/currency';
 import { type AddExpenseState, type Participant } from '~/store/addStore';
-import { getSupportedLanguages } from './i18n/client';
 
 export type ParametersExceptTranslation<F> = F extends (t: TFunction, ...rest: infer R) => any
   ? R
@@ -38,10 +37,12 @@ export const toUIDate = (
     }).format(date);
   }
 
-  return format(date, 'MMM dd', {
-    locale: getSupportedLanguages().find((lang) => lang.code === todayTranslation.usedLng)
-      ?.dateLocale,
-  });
+  const day = new Intl.DateTimeFormat(todayTranslation.usedLng, { day: '2-digit' }).format(date);
+  const monthName = new Intl.DateTimeFormat(todayTranslation.usedLng, { month: 'short' })
+    .format(date)
+    .replace('.', '');
+
+  return `${monthName} ${day}`;
 };
 
 export function generateSplitDescription(
--- a/src/utils/i18n/client.ts
+++ b/src/utils/i18n/client.ts
@@ -1,20 +1,17 @@
-import { type Locale, de, enUS, es, fr, it, pl, pt, ptBR, sv } from 'date-fns/locale';
-
 export interface SupportedLanguage {
   code: string;
   name: string;
-  dateLocale: Locale;
 }
 
 export const getSupportedLanguages = (): SupportedLanguage[] => [
-  { code: 'en', name: 'English', dateLocale: enUS },
-  { code: 'de', name: 'Deutsch', dateLocale: de },
-  { code: 'fr', name: 'Français', dateLocale: fr },
-  { code: 'it', name: 'Italiano', dateLocale: it },
-  { code: 'pl', name: 'Polski', dateLocale: pl },
-  { code: 'pt-PT', name: 'Portuguese (Portugal)', dateLocale: pt },
-  { code: 'pt-BR', name: 'Portuguese (Brazil)', dateLocale: ptBR },
-  { code: 'sv', name: 'Svenska', dateLocale: sv },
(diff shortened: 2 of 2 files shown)


src/utils/strings.ts
```diff
@@ -24,8 +25,24 @@ export const toUIDate = (
...
+  if (year) {
+    return Intl.DateTimeFormat(todayTranslation.usedLng, {
+      dateStyle: 'long',
+    }).format(date);
+  }
+
+  return format(date, 'MMM dd', {
+    locale: getSupportedLanguages().find((lang) => lang.code === todayTranslation.usedLng)
+      ?.dateLocale,
+  });
```
krokosik: I suppose we could just use `new Intl.DateTimeFormat(todayTranslation.usedLng, {day: '2-digit', month: 'short'}).format(date)`?

rodrigost23: I tried that, but this Intl API is kind of weird. At least for pt-BR and pt-PT, the results are awful, so I'm not sure what other languages would look like to native speakers.

### pt-PT:



I don't know why it would format "short" month to "11" (we have "numeric" for that, right?). 

### pt-BR:



The word "de" here is just a preposition that we can totally ignore when we want it to be shorter.

----

Maybe we could have a specific function only for this expense date prefix and use the Intl for other places, which would probably make this look better:

krokosik: I'm not a native speaker in either of these, but maybe it is simply the proper way to format this in Portugal? For the brazilian version, yeah it looks kind of bad.

Personally I would prefer fixing locale specific edge cases rather than to import the bulky `date-fns`  locale and pass it around.

For the expense list, we could format the day and month separately and remove dots for now?
```javascript
new Intl.DateTimeFormat('pt-PT', {month: 'short'}).format(new Date()).replace('.', '')
```
Actually returns the month name, so we could just built this in a consistent format. What do you think?

rodrigost23: Formatting day and month separately makes sense.

Do we need to care about the order, though? Some locales use MM/DD, others DD/MM.

<details><summary>There’s a way to detect that order if we really want to (expand for an example)</summary>
<table><tr><td>


```typescript
function formatDayMonth(locale: string, date: Date): string {
  // Format day and short month separately
  const day = new Intl.DateTimeFormat(locale, { day: 'numeric' }).format(date);
  const monthName = new Intl.DateTimeFormat(locale, { month: 'short' }).format(date)
    .replace('.', '');

  // Detect whether day comes before month for this locale
  const orderParts = new Intl.DateTimeFormat(locale, {
    day: 'numeric',
    month: 'numeric'
  }).formatToParts(date);

  const dayIndex = orderParts.findIndex(p => p.type === 'day');
  const monthIndex = orderParts.findIndex(p => p.type === 'month');

  const isDayFirst = dayIndex < monthIndex;

  return isDayFirst ? `${day} ${monthName}` : `${monthName} ${day}`;
}

// Examples:
formatDayMonth('pt-BR', new Date('2025-10-11'));
// "11 out"
formatDayMonth('pt-PT', new Date('2025-10-11'));
// "11 out"
formatDayMonth('de', new Date('2025-10-11'));
// "11 Okt" 
formatDayMonth('en-GB', new Date('2025-10-11'))
// "11 Oct" 
formatDayMonth('en-US', new Date('2025-10-11'));
// "Oct 11"
formatDayMonth('en', new Date('2025-10-11'));
// "Oct 11" 
formatDayMonth('zh', new Date('2025-10-11'))
// "10月 11日"
```

Feels a bit over-engineered, but it works.

In pt-BR I wouldn't mind if it stayed as "MM DD", since the layout stacks them on top of each other in the expense list.

Not sure how that reads in other locales... 

</td></tr></tr></table>
</details> 



What do you think? We just keep "MM DD" specifically in the expenses list for now? In that case, the following should be enough:

```typescript
const day = new Intl.DateTimeFormat(todayTranslation.usedLng, { day: 'numeric' }).format(date);
const monthName = new Intl.DateTimeFormat(todayTranslation.usedLng, { month: 'short' }).format(date).replace('.', '');

return `${monthName} ${day}`;

```

krokosik: Looks good to me. One change would be to change day to `2-digit` maybe?

krokosik: @rodrigost23 need any help with the PR?

rodrigost23: Sorry, just had been busy lately. I'll try finishing this one.

rodrigost23: I had the code already for this, so I just committed that now. I'm not totally satisfied with this just because it shows the same format in the date picker:
image

But the expenses list looks great:
image

krokosik: Would you prefer to make a separate date function just for the expenseList? That's fine by me since it is a bit hacky, while the `toUIDate` would rely on standard Intl API.

PR #472, opened 2025-11-05
Improve category picker layout

rodrigost23:
# Description
There were some issues in the category picker layout. It was all justified (resulting in weird spacing), and some localized texts could not fit (resulting in some text overlap):
image

This PR changes the layout to use a grid with a better responsiveness handling and adds text wrapping to the category names

# Demo

https://github.com/user-attachments/assets/8dd84047-3d90-484f-b946-6f6b063ec8d6

# Checklist


rodrigost23:
> Good fix, thank you! There is another thing that bothers me here: we have the `general` category listed twice. Could you add a fix for it to the PR?

That annoys me as well...
I removed the extra "general" subcategory and left only "other" in general. It seems this was the case before the i18n implementation.

image

PR #493, opened 2025-11-30
Feat 483 currency conversion polishing

### LINKED ISSUE #483: Currency conversion polishing
- [x] Handle notifications
- [x] make the backend procedure atomic, will be much easier with #481 
- [x] add balance view in preferred currency
- [x] use the new `CurrencyInput` component in conversions

krokosik:
# Description
Closes #483 

- notifications about currency conversions are now properly displayed
- the double expense for currency conversion is now added atomically
- adjusted the conversion logic for different decimal digits
- added possibility to convert balances on the fly

# Demo
https://github.com/user-attachments/assets/63b62b2b-be0f-4fe4-9be9-e30b5caaaf06

### DIFF
--- a/src/components/Expense/ConvertibleBalance.tsx
+++ b/src/components/Expense/ConvertibleBalance.tsx
@@ -195,7 +195,7 @@ export const ConvertibleBalance: React.FC<ConvertibleBalanceProps> = ({
           </span>
         ) : (
           <span className={0 < balances[0]!.amount ? 'text-emerald-500' : 'text-orange-600'}>
-            {getCurrencyHelpersCached(balances[0]!.currency).toUIString(balances[0]!.amount)}
+            {getCurrencyHelpersCached(balances[0]!.currency).toUIString(balances[0]!.amount)}*
           </span>
         )}
       </div>


pchampio:
Hello, loving the direction you are taking for split-pro.
I'm often reading the PRs/issues to see what's next and how I can help with issue reporting in between versions.
So as you said [here](https://github.com/oss-apps/split-pro/pull/479#issue-3605137836), I'm one of the brave peoples ;)

As such, when changes are made to any migration.sql file, like in this PR, the migration.sql are not re-run on our end, and we see catastrophic behavior during run-time.
When you make migration.sql files update that could be already run by people running main, could you update the date in the path too ? So that we see catastrophic behavior during launch-time.

krokosik:
@pchampio Yes, sorry about that. The migrations got really messy while working on this release and I'm trying to organized. Unfortunately, with this being the development build (and also an alpha version), I can not make any guarantees about nonbreaking Prisma migrations. To fix your database, you should change the column name in the expense table: `otherConversion` -> `conversionToId` and also update the foreign key.

I am working almost alone on this in my spare time and don't have a fleshed out plan for all the features. The beta release is when new code gets stabilized and I should have made it more explicit when posting the alpha release.

With that being said I'm glad you enjoy the app and direction! If everything has been working okay for you, could you also leave some feedback in the discussion? It's also helpful for me to know that stuff works rather than assuming nobody tested it :sweat_smile: 

Let me know if you need further assistance with fixing your DB

PR #506, opened 2025-12-31
Fix interpersonal balances affected by group debt simplification

### LINKED ISSUE #333: /balances page shows incorrect amounts after using Simplify Debts
**Describe the bug**
When a group is settled using the Simplify Debts feature, the /balances page does not show the correct amounts owed. Instead, it displays extra false debts.

**Example:**
Group of 3 people:

A: needs to receive money from B and C.
B: needs to receive money from C.
C: owes to A and B.

The simplify debts does reduce the number of "Settle up" transactions to make:

A: does not have to perform any transaction.
B: performs a transaction to A minus what C owes him/her.
C: performs a transaction to A plus what he/she owes to B.

However, on the `/balances` page it incorrectly shows:

A: owes money to C.
B: needs to receive money from C.
C: needs to receive money from A and owes the same amount to B.

**To Reproduce**

1. Create a group of 3 people.
2. Add expenses so that members owe each other.
3. Apply **Simplify Debts** and settle up.
4. Check the `/balances` page.

**Expected behavior**
The `/balances` page should reflect the simplified settlement:

**App version**
v1.5.0-beta.4
ChatGPT a dit :

This was a bit long to explain, feel free to ask questions, and I’ll try clarify.

krokosik:
# Description
Fixes #333 

We now specifically call the simplify algorithm when needed on affected groups. On the mock dataset which is probably larger than most instances the performance hit is not noticeable, but there is room for optimization if people need it.

Since the whole group/total balance intermingling is confusing I also expanded the UI to be more clear:
- friend settleups now also show balances from groups individually
- expense details page has been expanded with links to friends and groups if applicable. Also an indicator icon show if a group has simplify turned on.
- as a bonus, editing settleups now triggers the reopening of the settleup amount dialog rather than expense edit page, which actually broke the settlement db records

also added an `AGENTS.md` file as I started using opencode

# Demo
image
image

### DIFF
--- a/src/components/ui/button.tsx
+++ b/src/components/ui/button.tsx
@@ -34,8 +34,7 @@ const buttonVariants = cva(
 );
 
 export interface ButtonProps
-  extends React.ButtonHTMLAttributes<HTMLButtonElement>,
-    VariantProps<typeof buttonVariants> {
+  extends React.ButtonHTMLAttributes<HTMLButtonElement>, VariantProps<typeof buttonVariants> {
   asChild?: boolean;
   loading?: boolean;
   responsiveIcon?: boolean;
--- a/src/instrumentation.ts
+++ b/src/instrumentation.ts
@@ -11,9 +11,8 @@ export async function register() {
     const { validateAuthEnv } = await import('./server/auth');
     validateAuthEnv();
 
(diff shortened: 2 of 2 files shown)


laur89:
Good stuff!

PR #559, opened 2026-02-05
Fixes the amount input not updating after a currency conversion

pchampio:
After using the currency conversion dialog (e.g. USD → EUR), setAmount() correctly updated the internal BigInt value, but amountStr (the string displayed in the input field) was never updated. The user would see the old amount in the input even though the underlying value was already converted.

If there is a easier/cleaner fix, you can close the PR without comment.
Also, I haven't taken the time to setup a dev env (sorry, I cannot invest it at the moment, so it is very much docker build on my prod instance type of environment). As such, no linting/formatting is applied.


krokosik (approved):

krokosik:
Awesome, thank you for this 🙏

PR #566, opened 2026-02-10
Support GitLab as a generic OIDC provider

MaximilianHausen:
# Description
Adds a workaround for gitlab as an oicd provider. Gitlab returns a non-standard `created_at` field during authentication that needs to be discarded before passing it on to the database schema, just like what already happens for keycloak. Also adds a gitlab icon for the provider id "gitlab".

# Checklist

- [ ] The last commit successfully passed pre-commit checks (It should be fine, but I skipped the pre-commit checks because they weirdly capitalized an unrelated multi-line comment later in the file)

### DIFF
--- a/src/server/auth.ts
+++ b/src/server/auth.ts
@@ -76,17 +76,20 @@ const SplitProPrismaAdapter = (...args: Parameters<typeof PrismaAdapter>): Adapt
         throw new Error('Adapter is missing the linkAccount method.');
       }
 
+      // Keycloak and Gitlab provide some non-standard fields that do not exist in the prisma schema.
+      // We strip them out before passing them on to the original adapter.
       if (account.provider === 'keycloak') {
-        // Keycloak provides some non-standard fields that do not exist in the prisma schema.
-        // We strip them out before passing them on to the original adapter.
         const {
-          ['not-before-policy']: _notBeforePolicy,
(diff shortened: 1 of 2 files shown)


src/server/auth.ts
```diff
@@ -76,17 +76,20 @@ const SplitProPrismaAdapter = (...args: Parameters<typeof PrismaAdapter>): Adapt
...
       }
 
+      // Keycloak and Gitlab provide some non-standard fields that do not exist in the prisma schema.
+      // We strip them out before passing them on to the original adapter.
       if (account.provider === 'keycloak') {
-        // Keycloak provides some non-standard fields that do not exist in the prisma schema.
-        // We strip them out before passing them on to the original adapter.
         const {
-          ['not-before-policy']: _notBeforePolicy,
+          'not-before-policy': _notBeforePolicy,
```
krokosik, line 83: Why is Keycloak logic being modified? I don't have a Keycloak instance so I have no way of testing if it breaks.

MaximilianHausen: I changed it because it triggered a no-useless-computed-key linter warning, but didn't verify it further.

MaximilianHausen: I’ve now tested the keycloak sign-in with a local instance, and it still works.

krokosik: cool, thanks

PR #574, opened 2026-02-13
Display archived groups when no active groups exist

timstott:
# Description

The list of archived groups isn't displayed on the groups page when there are no active groups.

🤖 I’ve added tests to verify that expected elements render correctly in the DOM. I used AI to generate the initial boilerplate, as I’m not yet familiar with writing tests in the Next.js framework. There’s also a question of whether Cypress would be preferable to introduce end-to-end tests for a broader regression prevention.

# Demo

| Before | After |
|--------|--------|
| Screenshot 2026-02-13 at 08-03-30 Groups | Screenshot 2026-02-16 at 07-25-54 Groups |

# Checklist

### DIFF
--- a/src/pages/groups.tsx
+++ b/src/pages/groups.tsx
@@ -45,7 +45,7 @@ const BalancePage: NextPageWithUser = () => {
       <MainLayout title={t('navigation.groups')} actions={actions} loading={groupQuery.isPending}>
         <div className="mt-7 flex flex-col gap-8 pb-36">
           {0 === groupQuery.data?.length ? (
-            <div className="mt-[30vh] flex flex-col items-center justify-center gap-20">
+            <div className="mt-[30vh] mb-16 flex flex-col items-center justify-center gap-20">
               <CreateGroup>
                 <Button>
                   <PlusIcon className="mr-2 h-4 w-4" />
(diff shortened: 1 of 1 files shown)


src/pages/groups.tsx
```diff
@@ -44,7 +47,7 @@ const BalancePage: NextPageWithUser = () => {
       </Head>
       <MainLayout title={t('navigation.groups')} actions={actions} loading={groupQuery.isPending}>
         <div className="mt-7 flex flex-col gap-8 pb-36">
-          {0 === groupQuery.data?.length ? (
```
krokosik, line 47: I think it's more sensible to just move the archived groups out of the ternary operator,.

src/tests/pages/groups.test.tsx
```diff
@@ -0,0 +1,137 @@
+import '@testing-library/jest-dom';
+
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+
+import BalancePage from '~/pages/groups';
+
+// Next-i18next – return translation keys as-is
+jest.mock('next-i18next', () => ({
```
krokosik: I haven't got the time yet to get into testing Next components, but if it's supposed to look like this, then unfortunately I'm not interested. There must be a saner way to manage all these mocks so as it is readable by a human=. As such, don't worry about tests in your PR and please remove these changes.

krokosik (changes requested):

src/tests/pages/groups.test.tsx
timstott: Agreed, it's currently a bit clunky. This can be improved with a wrapper function that mounts the application but the data layer would still need to be mocked/stubeed. What are your thoughts on the testing approach for the project, would you rather see end-to-end tests with something like Cypress or use Jest DOM with an improved interface?

Tests removed ✅ .

src/pages/groups.tsx
timstott: Done, I've also added a bit of bottom margin to the create button component to give some breathing room with the Archived accordion.

Screenshop updated in PR description.

krokosik:
Sorry for the delayed review. The UI is a bit weird when there are only archived groups left. Maybe it would make more sense to move the accordion to the bottom instead?

timstott:
Hi @krokosik. I apologise for the late reply. Feel free to take over the fix to save us from iterating over the placement of the UI element.

PR #591, opened 2026-03-20
Correctly format amounts of less than 10 cents

FriesischScott:
# Description

Amounts of less than 10 cents are currently formatted incorrectly. For example 1 cent (represented as `1n`) is rendered as `$0.1`. 

When `simplifyDebts` results in very small amounts they render incorrectly and do not add up to the correct debts anymore.

The issue is fixed by padding the string with the appropriate number of zeros.

I've also added a corresponding test.

# Checklist


krokosik:
Nice catch! Thanks

PR #637, opened 2026-05-04
Enable filtering currencies by name.

### LINKED ISSUE #636: Enable filtering by currency name
**Is your feature request related to a problem? Maybe it would be better as discussion? Please describe.**
I was annoyed to have to type the currency code instead of the name when switching between currencies, e.g. "Yen" is not recognized so you have to type "JPY".

When typing a string in the search box of the CurrencyPicker it should also be used to filter based on the currency name.

**Describe alternatives you've considered**
With #600 merged the default behavior should be okay since I'm only switching between a handful of currencies regularly. So this is not really a problem anymore but maybe it's still nice to have.

**App version**
Not supported in latest main branch.

addap:
Fixes #636.

# Description
Use the keywords property of CommandItem which can receive extra arguments to filter for.

# Demo
splitpro

# Checklist

- [x] Any AI code was thoroughly reviewed by me (none)

### DIFF
--- a/src/components/AddExpense/CurrencyPicker.tsx
+++ b/src/components/AddExpense/CurrencyPicker.tsx
@@ -76,6 +76,17 @@ function CurrencyPickerInner({
     [],
   );
   const extractKey = extractValue;
+  const extractKeywords = useCallback(
+    (currency: { code: CurrencyCode } | { code: '__CLEAR__' }) => {
+      if ('__CLEAR__' === currency.code) {
+        return [];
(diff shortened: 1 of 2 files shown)


addap:
oops sorry, I ran the pre-commit check but did not notice it actually reformatted something which did not end up in the commit. I fixed that and ran the `check` workflow in my fork. Should work now.

krokosik:
Awesome improvement, thank you!

PR #647, opened 2026-05-17
Fix #631: respect user and group default currency when creating expenses

### LINKED ISSUE #631: [Bug] Unable to set default currency
**Describe the bug**
When adding an expense, it always defaults to USD. Both my personal account (default currency), and the group default is JPY. If i log out (or close browser), it reverts back to USD. Occurs on mobile and PC.

**To Reproduce**
1. Open expense 
2. Change from USD to JPY
3. Add a cost, press ok.
4. Create a new expense. Stays as JPY
5. Close window (or log out).
6. Create a new expense - reverts to USD.

**Expected behavior**
If the Group and the user have default JPY, creating an expense (either direct, or in group) i would expect it to default as JPY

**Screenshots**

Image

Image

Image

<close window>
<add expense>

Image

Proof:

Image

Image

**App version**
v2.1.0

Debug Information

**Desktop (please complete the following information):**

:UserAgentMozilla/5.0 (Windows NT 10.0; Win64; x64; rv:150.0) Gecko/20100101 Firefox/150.0Git1be122ff0c6a3864df1bc12237379208177d0667Versionv2.1.0
Edition	Windows 11 Pro
Version	25H2
Installed on	‎10/‎11/‎2025
OS build	26200.7171
Experience	Windows Feature Experience Pack 1000.26100.265.0

**Smartphone (please complete the following information):**
 - Device: Xiaimio Redmi note 13 pro 5g
 - OS: xiaomi hyper os 2.0.206.0
 - Android version 15 Aq3A
UserAgentMozilla/5.0 (Android 15; Mobile; rv:150.0) Gecko/150.0 Firefox/150.0Git1be122ff0c6a3864df1bc12237379208177d0667Versionv2.1.0

**Additional context**
important note:
- This is hosted on a raspberry pi 3b using docker.
- The login/invite has been disabled.
- I'm not super tech savvy, but have wiped the sd card several times, reproduced each time.
- As i got a bit frustrated, i've used AI for assistance, which has probably made things worse (trying to force JPY settings in the environment/back end). So will need to wipe it all again. However i think this is a bug, as i wouldnt have tried to force

phof:
## Summary

Closes #631.

The expense form always defaulted to USD regardless of the user's or group's configured default currency. The defaults were fetched from the backend but never applied to the form's currency field.

## Changes

- **`src/pages/add.tsx`**:
  - For non-group expenses, initializes the currency from \`user.defaultCurrency\` in the \`setCurrentUser\` effect.
  - For group expenses, initializes the currency from \`group.defaultCurrency\` (falling back to the user's default) in the group initialization effect.
  - The \`setCurrentUser\` effect is gated on \`router.isReady\` so that the first render — before Next.js populates \`router.query\` — does not falsely treat a group page as a non-group page and apply the wrong default.
  - Moved the \`useRouter\` / \`router.query\` destructuring above the \`setCurrentUser\` effect so \`groupId\` is in scope.
- **\`src/store/addStore.ts\`**: \`resetState\` now restores the user's default currency on unmount, so navigating away and back lands on the correct default instead of stale state.

## Test plan

- [x] Open \`/add\` as a user whose default currency is set (e.g. JPY). Currency picker shows JPY.
- [x] Open \`/add?groupId=N\` for a group whose default currency is EUR. Currency picker shows EUR.
- [x] Open \`/add?groupId=N\` for a group with no default currency, where the user default is JPY. Picker shows JPY.
- [x] Manually change currency to USD on the add page, navigate away, navigate back. Picker resets to the user's default.
- [x] Close the window and reopen. Picker still shows the user's default (not USD).
- [x] Edit an existing expense: picker still shows the expense's original currency (unchanged behaviour).

## Disclaimer

Parts of this change were drafted with the help of Claude Opus 4.7.

### DIFF
--- a/src/store/addStore.ts
+++ b/src/store/addStore.ts
@@ -3,7 +3,7 @@ import Router from 'next/router';
 import { create } from 'zustand';
 
 import { DEFAULT_CATEGORY } from '~/lib/category';
-import { type CurrencyCode, isCurrencyCode } from '~/lib/currency';
+import { type CurrencyCode } from '~/lib/currency';
 import type { TransactionAddInputModel } from '~/types';
 import { shuffleArray } from '~/utils/array';
 import { BigMath } from '~/utils/numbers';
@@ -243,10 +243,6 @@ export const useAddExpenseStore = create<AddExpenseState>()((set) => ({
         cronExpression: '',
         isFileUploading: false,
         paidBy: s.currentUser,
-        currency:
-          s.currentUser?.defaultCurrency && isCurrencyCode(s.currentUser.defaultCurrency)
-            ? s.currentUser.defaultCurrency
-            : s.currency,
       }));
     },
     setSplitScreenOpen: (splitScreenOpen) => set({ splitScreenOpen }),
--- a/src/store/addStore.ts
+++ b/src/store/addStore.ts
@@ -3,7 +3,7 @@ import Router from 'next/router';
 import { create } from 'zustand';
 
 import { DEFAULT_CATEGORY } from '~/lib/category';
-import { type CurrencyCode } from '~/lib/currency';
+import { type CurrencyCode, isCurrencyCode } from '~/lib/currency';
 import type { TransactionAddInputModel } from '~/types';
 import { shuffleArray } from '~/utils/array';
 import { BigMath } from '~/utils/numbers';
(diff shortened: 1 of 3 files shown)


src/store/addStore.ts
```diff
@@ -243,6 +243,10 @@ export const useAddExpenseStore = create<AddExpenseState>()((set) => ({
         cronExpression: '',
         isFileUploading: false,
         paidBy: s.currentUser,
+        currency:
```
krokosik: Please remove this. The core functionality that suffered a regression is to keep the last used currency, so the reset function should not change anything in the currency field

krokosik (changes requested):
The fix is appropriate but not quite there yet. It omits the core of the regression, which is the missing functionality of defaulting to last used currency. It is kept in `currency` field in the `currentUser` object. The correct preference chain is:
1. `currentUser.currency`
2. `group.defaultCurrency`
3. `currentUser.defaultCurrency`

phof:
@krokosik Thank you, see if the latest changes make sense now

krokosik (approved):

PR #465, opened 2025-11-04
Import transactions from Splitwise

rodrigost23:
# Description

This is an idea that would require https://github.com/oss-apps/export-splitwise/pull/2 to be merged; meaning the user would still export a JSON from that external website, then import to Split Pro.

It isn't so efficient as the one proposed in #307, but it did work with 1600+ transactions pretty quick on my computer. Not sure if this would be a problem.

Closes #307

# Demo

This PR is WIP and still needs more testing. I want to make a json with some dummy data to test and to have screenshots/videos to post here.

# Checklist

- [ ] I have read `CONTRIBUTING.md` in its entirety
- [x] I have performed a self-review of my own code
- [ ] I have added unit tests to cover my changes
- [x] The last commit successfully passed pre-commit checks
- [x] Any AI code was thoroughly reviewed by me

PR #660, opened 2026-05-24
feat(import): add Splitwise API import for full expense history

### LINKED ISSUE #307: Splitwise import 2.0
## The Context

There are numerous issues raised (#271 #207 #46 ...) with the current state of Splitwise import and I would like to group them all together, as well as gather feedback for how I would like to rework the current flow.

Currently, we rely on an external hosted solution, to generate a JSON with SOME data fetched from Splitwise API. Looking at the code, it is very [simple](https://github.com/oss-apps/export-splitwise/blob/main/src/app/api/download/route.ts). We do not import expenses and with the planned #234, such imported accounts will break.

The background for not importing expenses was the fact, that we do not want to overload the Node server with processing of JSONs, which could slow down the app (this concern was probably related more to the public instance) #29 

## The Plan

- [ ] Add `pg-boss`, which will allow us to offload execution to another process. We also want this for #79 and #160 
- [ ] Change the import splitwise screen, to have an accordion from https://export-splitwise.vercel.app/ and accept the API key not JSON
- [ ] Create a worker job that first does exactly what the current functionality does
- [ ] Fetch ALL transactions in batches
- [ ] Fetch more transaction metadata

Along the way, we might switch to https://github.com/athulanilthomas/splitwise-ts

### Nice to have

- [ ] more granular import, allow importing a subset of groups/friends
- [ ] toggle for archived groups
- [ ] toggle for not importing settled up expenses

Mufaddal5253110:
# Description

  Adds direct Splitwise API import for complete expense history with individual transactions.

  **What it does:**
  - Imports individual expenses (not just net balances) with correct `paid_share` / `owed_share` calculations
  - Preserves original `createdAt` timestamps for proper Activity ordering
  - Uses `transactionId` to prevent duplicates (safe to re-run)

  **Files added:**
  - `src/lib/splitwise-api.ts` - API client with pagination
  - `src/lib/splitwise-converter.ts` - Conversion logic
  - `src/server/api/routers/splitwiseApiImport.ts` - tRPC endpoints
  - `src/pages/import-splitwise-api.tsx` - UI for API key entry and group selection
  - `src/tests/splitwise-converter.test.ts` - 29 test cases

  Closes #307
  Related to #465 (different approach - direct API vs external tool)

  ## ⚠️ ToS Discussion Needed

  As noted in #307, Splitwise's API ToS prohibits use for competing services. This PR uses the direct API because the JSON export from export-splitwise only contains **net balances**, not individual expenses.

  **Options to discuss:**
  1. Keep API import with disclaimer (user provides their own key)
  2. Remove and rely on CSV/JSON import only
  3. Other suggestions?

  # Demo

  1. User enters Splitwise API key (from secure.splitwise.com/apps)
  2. Selects groups to import
  3. Expenses imported with full history

  # Checklist

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -31,6 +31,7 @@
     },
     "follow_on_x": "Follow us on X",
     "import_from_splitwise": "Import from Splitwise",
+    "import_from_splitwise_api": "Import from Splitwise (API)",
     "import_from_splitwise_details": {
       "choose_file": "Choose file",
       "export_splitwise_data_button": "Export splitwise data",
--- a/src/lib/splitwise-api.ts
+++ b/src/lib/splitwise-api.ts
@@ -1,0 +1,304 @@
+/**
+ * Splitwise API Service
+ *
+ * Provides methods to fetch data from Splitwise API for importing
+ * expenses into SplitPro with complete accuracy.
+ */
+
+const API_BASE = 'https://secure.splitwise.com/api/v3.0';
+
+// ============= TYPES =============
+
+export interface SplitwiseUserShare {
+  user_id: number;
+  user: {
+    id: number;
+    first_name: string;
+    last_name: string | null;
+    email: string;
+    picture?: {
+      medium?: string;
+    };
+  };
+  paid_share: string;
+  owed_share: string;
+  net_balance: string;
+}
+
+export interface SplitwiseExpense {
+  id: number;
+  cost: string;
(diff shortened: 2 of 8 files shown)


krokosik:
Sorry, I don't accept 100% vibe coded PRs and this is clearly one of them. Most importantly I cannot accept it due to the fact that it violates the Splitwise API ToS.

PR #662, opened 2026-05-25
fix: prevent user/group ID collision in currency preference store

### LINKED ISSUE #661: Balance displays in USD instead of EUR for specific user after tab navigation
Describe the bug
The balance for a specific user (degrooteglenn) displays in USD ($) instead of EUR (€) after navigating away from and back to the Balances page. On initial load the correct EUR amount appears, but after any tab navigation the value switches to USD and remains there persistently. All other users display correctly in EUR throughout.

To Reproduce

	1.	Open the Balances page → degrooteglenn shows €286.87
	2.	Navigate to another tab (e.g. Groups)
	3.	Navigate back to Balances
	4.	degrooteglenn now shows ≈ $332.63
	5.	The USD display persists for this user across all further navigation

Expected behavior
All balances display in EUR at all times. No USD conversion should occur when no USD transactions exist and the affected user has EUR configured as their currency.

Screenshots
(two screenshots attached — initial load showing €286.87, post-navigation showing ≈$332.63 for the same user at the same timestamp 23:22)

First load
Image

Returning after different tab:

Image

App version
v2.1.2 — commit 6dc90b437740c24f36fc40db36b0cd78f1f0a071

Smartphone:

	•	Device: iPhone
	•	OS: iOS 18.7
	•	Browser: Safari Mobile — WebKit 605.1.15 / Version 26.4
	•	Version: v2.1.2

Additional context
The affected user has EUR set as their currency. No payments in USD are visible anywhere in the transaction history. The ≈ symbol indicates the app is performing a currency conversion. The bug is navigation-triggered: correct EUR is shown on initial mount, USD appears after the component re-renders following tab navigation. This points to a client-side state or cache issue where the currency for this specific user is incorrectly resolved on re-render — possibly stale or incorrectly keyed currency state that gets populated on the second fetch but not on initial mount.

jaredslorenz:
# Description
Closes #661
Integer ID collision between user IDs and group IDs in currencyPreferenceStore caused friend balances to display in the wrong currency after tab navigation. The getPreference function's fallback chain checked groupDefaultCurrencies for any numeric key — so if a user's ID matched a group ID with USD as its default, navigating to the Groups tab would poison the store and convert that user's EUR balance to USD on re-render.
Changes:

currencyPreferenceStore.ts — getPreference now takes entityType?: 'group'; group currencies only consulted when entityType === 'group'
ConvertibleBalance.tsx — Added entityType?: 'group' prop; gates group currency lookup accordingly
CumulatedBalances.tsx — Forwards entityType prop through to ConvertibleBalance
GroupMyBalance.tsx — Passes entityType="group" as the only correct caller for group currency defaulting

AI assistance was used to help diagnose the root cause and implement the fix. All changes were thoroughly reviewed by me before submission.
# Demo
Unable to reproduce locally without the specific user data, but the fix is verified by code analysis and unit tests covering the exact collision scenario.
# Checklist

 I have read CONTRIBUTING.md in its entirety
 I have performed a self-review of my own code
 I have added unit tests to cover my changes
 The last commit successfully passed pre-commit checks
 Any AI code was thoroughly reviewed by me


krokosik (commented):
Very nice, but what happened to the .gitignore file?

krokosik:
Re: the `.gitignore` change — it was an accidental line-ending change (CRLF) from a Windows machine, which made git treat it as a binary file with no actual content change. I've reverted it to match `main` exactly.

Also fixed a prettier formatting issue in `ConvertibleBalance.tsx` that was causing the CI check to fail. Both fixes are in commit `8040560`.

The fix and tests are unchanged — great work on identifying the ID collision root cause! 🎯

jaredslorenz:
Thanks for catching that and cleaning it up! I appreciate the review.

Issue #418, opened 2025-10-05
i18n: Use full sentences for translation keys instead of fragments


rodrigost23:
**Description**

The current i18n implementation composes sentences from multiple, separate translation keys. This is an internationalization (i18n) anti-pattern that prevents accurate and natural-sounding translations for languages with different grammatical structures.

-----

### **Current Implementation Examples**

The following code snippets show how sentence fragments are currently being concatenated directly in the UI components:

**1. Hero Section Titles (`/components/landing/Hero.tsx`)**
The main title and subtitle are constructed from different keys.

```jsx
              <h1 ...>
                {t('hero.title_part1')}{' '}
                <span ...>{t('hero.title_highlight')}</span>.
              </h1>
              <h2 ...>
                {t('hero.subtitle_part1')}{' '}
                <a ...>
                  {t('hero.subtitle_link')}
                </a>{' '}
                {t('hero.subtitle_part2')}
              </h2>
```

**2. Expense Details (`/components/expense/ExpenseDetails.tsx`)**
A sentence is formed by injecting a translated preposition between two variables.

```jsx
{toUIString(e.amount)} {t('ui.expense.to')} {displayName(userDetails.data, userId)}
```

-----

### **Why This Is a Problem**

- **Incorrect Grammar:** It hardcodes the English sentence structure, which will break translations for languages with different word orders.
- **Ambiguous Translations:** It forces a single translation for words that require context. For example, the prepositions **`"for"`** and **`"to"`** (`ui.expense.for`, `ui.expense.to`) can have multiple, different translations in other languages depending on the sentence's meaning (e.g., purpose, direction, recipient). In some cases, they might even translate to the same word. Without the full sentence, a translator cannot choose the correct form.

-----

### **Expected Behavior **

Translation keys should contain complete sentences, using interpolation for dynamic values. This gives translators the full context they need to create a correct and natural translation.

**Example Refactoring:**

**Before:**

```json
  "ui": {
    ...
    "expense": {
      "for": "for",
		...
      "to": "to",
		...
	}
	...
}
```

```jsx
{toUIString(e.amount)} {t('ui.expense.to')} {displayName(userDetails.data, userId)}
```

**After:**

```json
"ui": {
  "expense": {
    "amount_owed_to_user": "{{amount}} to {{name}}"
  }
}
```
```jsx
// component.jsx
t('ui.expense.amount_owed_to_user', {
  amount: toUIString(e.amount),
  name: displayName(userDetails.data, userId)
})
[...]

PR #669, opened 2026-05-30
Merging user.owe & actors.you_dativus into user.owe_to_you

NeilSCGH:
# Description
In this PR, I'm creating a new `ui.expense.user.owe_to_you` translation, by merging `ui.expense.user.owe` & `actors.you_dativus`, and updating `/src/components/group/GroupMyBalance.tsx` to use this new field.

This is used to fix a mistake in the french translation:
`ui.expense.user.owe` & `actors.you_dativus` was giving "dois tu" but in french we should say "te doit".

In order to not mess up the other languages, I've manually created `owe_to_you` in all of them by merging user.owe + you_davitus, so it should not change anything for them.

# Checklist


krokosik:
Well, I can already see that it breaks the Polish translation. I am more in favor of a more robust solution using ICU, please add this example to #418

PR #676, opened 2026-06-12
fix: clearer unequal-split language and prevent silent save failure

### LINKED ISSUE #645: Improve confusing language for unequal splits
The flow for adding an expense where 1 person is owed the full amount is confusing/error-prone.

To reproduce:

1. Add expense. Description = Test, Amount = $1
2. Hit "split equally (2)" to go to the split option. 

At this screen, it's unclear if the amounts mean `you owe $` vs `you are owed $`. I mistakenly assumed `you are owed $` (fwiw, my friend did the same thing last week when she tried to the same thing).

Image

If you click "You", the screen shows

Image

"paid for You" is ambiguous here. If you hit "Save", the transaction silently fails and never shows up.

Clearer language to show the direction where money is owed. fwiw, Splitwise is pretty unambiguous and provides "50:50" or "full amount" as defaults, which is sensible

Image

**Describe alternatives you've considered**

**App version**
2.1.1

**Additional context**

mvanhorn:
# Description

Scopes the fix for #645 to the two things @krokosik was receptive to in the thread: the silent save failure and the ambiguous "paid for/by you" wording.

When a split left one person owed the full amount, the secondary screen's Save could no-op and the expense would silently disappear. This adds a blocking validation guard in `addStore` so Save is prevented with an inline message instead of discarding the expense, and rewords the directional labels so the direction of money is explicit. Split-definition semantics are unchanged — this only clarifies labels and adds the missing guard, staying inside the scope @krokosik agreed to (no rework of the core split modal to show money flows).

Fixes #645

# Demo

Non-visual validation + wording change, covered by unit tests. `src/tests/addStore.test.ts` adds cases for the one-person-owed split's `canSplitScreenClosed`/validity flag and the adjustment-exceeds-total guard; the previously-vanishing expense now blocks Save with a validation error. `SKIP_ENV_VALIDATION=true pnpm test` passes (49/49). No jest config change is needed — CI's `check.yml` already exports `SKIP_ENV_VALIDATION`.

# Checklist

## Summary by CodeRabbit

* **New Features**
  * Added clearer debt-direction labels for expense splits, including who owes whom.
  * Added localized guidance when an expense split is invalid or incomplete.

* **Bug Fixes**
  * Improved equal-split validation to prevent closing splits with no valid amounts.
  * Corrected split descriptions for negative balances, single participants, and no-money-flow scenarios.

* **Tests**
  * Added coverage for incomplete and single-participant equal-split scenarios.

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -189,6 +189,12 @@
         "warning": "Warning: Don't use send invite if it's invalid email. use add to Split Pro instead. Your account will be blocked if this feature is misused"
       },
       "split_type_section": {
+        "direction": {
+          "no_money_flow": "No one else owes anything",
+          "owes_payer": "{{debtor}} owes {{payer}}",
+          "owes_you": "{{debtor}} owes you",
+          "you_owe": "You owe {{payer}}"
+        },
         "split_equally": "split equally",
         "split_unequally": "split unequally",
         "types": {
@@ -213,6 +219,9 @@
             "title": "Share",
             "total_shares": "Total shares"
           }
+        },
+        "validation": {
+          "invalid_split": "Adjust who owes this expense before saving."
         }
       },
       "sponsor_us": "Sponsor us",
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -137,6 +137,7 @@
     "group_name_update_failed": "Error in updating group name",
     "import_failed": "Error importing file",
     "invalid_currency_code": "Invalid currency code {{code}}",
+    "invalid_split": "Adjust who owes this expense before saving.",
     "language_change_failed": "Error changing language",
     "less_than": "File should be less than {{size}}MB",
     "image_compression_failed": "Image compression failed, trying to upload original file",
@@ -219,9 +220,6 @@
             "title": "Share",
             "total_shares": "Total shares"
           }
-        },
-        "validation": {
-          "invalid_split": "Adjust who owes this expense before saving."
         }
       },
       "sponsor_us": "Sponsor us",
--- a/src/components/AddExpense/AddExpensePage.tsx
+++ b/src/components/AddExpense/AddExpensePage.tsx
@@ -1,3 +1,4 @@
+import { SplitType } from '@prisma/client';
 import { HeartHandshakeIcon, Landmark, RefreshCcwDot, X } from 'lucide-react';
 import { useTranslation } from 'next-i18next';
 import Link from 'next/link';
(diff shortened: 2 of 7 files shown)


src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -55,8 +56,73 @@ export const AddOrEditExpensePage: React.FC<{
...
+      }
+
+      return t('expense_details.add_expense_details.split_type_section.direction.owes_payer', {
+        debtor: debtorName,
+        payer: payerName,
+      });
+    }
+
+    return `${splitEquallyText} (${selectedParticipants.length})`;
+  }, [
```
krokosik: There is already a util function encapsulating this behavior `generateSplitDescription`. I would prefer for such extended logic to be kept out of the already huge AddExpensePage file.

src/components/AddExpense/AddExpensePage.tsx
```diff
@@ -345,16 +413,13 @@ export const AddOrEditExpensePage: React.FC<{
...
-                        participants,
-                        splitShares,
-                        paidBy,
-                        currentUser,
-                      )}
+                      {splitDescription}
                     </Button>
                   </SplitExpenseForm>
                 </div>
+                {!isExpenseSettled ? (
```
krokosik: The validation should already be handled in `splitTypeSection`. There should be no way of finding oneself outside of that input with an invalid state, so this logic is redundant, please remove it along with the toast.

src/store/addStore.ts
```diff
@@ -294,11 +294,12 @@ export function calculateParticipantSplit(
       const totalParticipants = participants.filter((p) => 0n !== getSplitShare(p)).length;
       updatedParticipants = participants.map((p) => ({
         ...p,
-        amount: 0n === getSplitShare(p) ? 0n : amount / BigInt(totalParticipants),
+        amount:
+          0 === totalParticipants || 0n === getSplitShare(p)
```
krokosik: See how `totalParticipants` is calculated. It's the amount of non-zero splitshares, so this check is completely redundant and complicates the most sensitivie part of the code

src/store/addStore.ts
```diff
@@ -371,6 +379,14 @@ export function calculateParticipantSplit(
...
       }
     }
+
+    if (
+      canSplitScreenClosed &&
+      1 < participants.length &&
+      updatedParticipants.every((p) => 0n === (p.amount ?? 0n))
+    ) {
+      canSplitScreenClosed = false;
+    }
```
krokosik, line 379: Care to explain why do you need to refactor such a huge chunk of the core logic? AFAIK all you want is to check an additional edge case, where all the participant.amount fields are zero.

krokosik (changes requested):

mvanhorn:
Pushed the requested changes in caa4f29. Lint, type-check, the full 141-test suite, and a production build all pass. Ready for another look.

public/locales/en/common.json
```diff
@@ -213,6 +219,9 @@
             "title": "Share",
             "total_shares": "Total shares"
           }
+        },
+        "validation": {
+          "invalid_split": "Adjust who owes this expense before saving."
```
krokosik: move this to errors

krokosik (changes requested):

mvanhorn:
Moved it to `errors.invalid_split` in dd70674 and updated the call site. No other locale carried the old key, so nothing else needed touching.

krokosik:
One more thing that could be added to this PR is disabling the Save button when the expense is not properly settled. We are good to go then :)

mvanhorn:
Save is now disabled while the split is unsettled, using the same `isExpenseSettled` guard the bottom action already had. Also gave the invalid-split message `role="alert"` so it gets announced when it appears.

mvanhorn:
Thanks for merging the unequal-split language fix, @krokosik. Fewer silent surprises in shared expenses.

krokosik:
Thank you for your work :)

PR #682, opened 2026-06-17
feat: add Kanidm OIDC support with PKCE and ID token algo config

mikimandoki:
# Description

While I was attempting to set up [Kanidm](https://github.com/kanidm/kanidm) to work with SplitPro I had hit a few roadblocks: 

1. Kanidm includes an `issued_token_type` field in its token response which isn't in the Prisma schema, causing a DB crash before the account could be created. Rather than adding another provider-specific case to `linkAccount`, this PR replaces the entire chain with a generic sanitisation pass using `Prisma.AccountScalarFieldEnum` as the source of truth. Any provider with non-standard fields now works without further changes.

2. Kanidm also enforces PKCE and ES256 by default, both of which were previously not supported by SplitPro. These can now be configured by optional env vars. I also added an example kanidm env config. 

To test these changes, I ran the updated code on my NAS in a Docker container and finally managed to authenticate with Kanidm. Disabling PKCE or defaulting to RS256 both fail on the default Kanidm config. 

# Checklist

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -146,20 +146,6 @@ OIDC_CLIENT_SECRET="<client-secret>"
 OIDC_WELL_KNOWN_URL="https://keycloak.example.com/realms/My_Realm/.well-known/openid-configuration"
 ```
 
-### OIDC (Kanidm)
-
-```bash
-DATABASE_URL="postgresql://postgres:strong-password@localhost:5432/splitpro"
-NEXTAUTH_SECRET="<generated>"
-NEXTAUTH_URL="https://splitpro.example.com"
-OIDC_NAME="kanidm"
-OIDC_CLIENT_ID="<client-id>"
-OIDC_CLIENT_SECRET="<client-secret>"
-OIDC_WELL_KNOWN_URL="https://kanidm.example.com/oauth2/openid/<app-name>/.well-known/openid-configuration"
-OIDC_PKCE_ENABLED=1
-OIDC_ID_TOKEN_SIGNED_RESPONSE_ALG=ES256
-```
-
 ## Security notes
 
 - Rotate `NEXTAUTH_SECRET` if it is ever exposed.
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -91,6 +91,8 @@ Used for magic-link login and invites.
 - `OIDC_CLIENT_SECRET`
 - `OIDC_WELL_KNOWN_URL`: OpenID well-known discovery URL.
 - `OIDC_ALLOW_DANGEROUS_EMAIL_LINKING`: Optional flag to allow email-based account linking.
+- `OIDC_PKCE_ENABLED`: Optional flag to enable PKCE (Proof Key for Code Exchange).
+- `OIDC_ID_TOKEN_SIGNED_RESPONSE_ALG`: Optional ID token signed response algorithm (e.g. `ES256`, `RS256`).
 
 ### Web push notifications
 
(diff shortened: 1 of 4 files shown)


src/server/auth.ts
```diff
@@ -253,6 +243,10 @@ function getProviders() {
       type: 'oauth',
       wellKnown: env.OIDC_WELL_KNOWN_URL,
       authorization: { params: { scope: 'openid email profile' } },
+      checks: env.OIDC_PKCE_ENABLED ? ['pkce'] : [],
```
krokosik, line 246: As far as I understand OIDC, the encryption configs should be provided from the `well known` url endpoint. I am against opening up the env var config to the whole breadth of OAuth2 options as this would just continue increasing and be a bad incentive for AI agents.

I am not familiar with Kandim, but please double check it's not an issue on their end or with your configuration, as setting these options by hand should not be required.

src/server/auth.ts
```diff
@@ -71,32 +72,21 @@ const SplitProPrismaAdapter = (...args: Parameters<typeof PrismaAdapter>): Adapt
...
-      // Keycloak and Gitlab provide some non-standard fields that do not exist in the prisma schema.
+      // OIDC providers can provide non-standard fields that do not exist in the prisma schema.
       // We strip them out before passing them on to the original adapter.
-      if (account.provider === 'keycloak') {
-        const {
-          'not-before-policy': _notBeforePolicy,
-          refresh_expires_in: _refresh_expires_in,
-          ...standardAccountData
-        } = account as AdapterAccount & Record<string, unknown>;
+      const knownAccountFields = new Set<string>(Object.values(Prisma.AccountScalarFieldEnum));
```
krokosik, line 81: very cool

docs/CONFIGURATION.md
```diff
@@ -144,6 +146,20 @@ OIDC_CLIENT_SECRET="<client-secret>"
 OIDC_WELL_KNOWN_URL="https://keycloak.example.com/realms/My_Realm/.well-known/openid-configuration"
 ```
 
+### OIDC (Kanidm)
```
krokosik: This document should be concise and does not have space for the myriad of OAuth providers out there. Respective values should be available in the provider's docs, please remove this.

src/server/auth.ts
mikimandoki: Thanks for the feedback. I'm no OIDC expert myself. Here's what I could find out: 

Kanidm correctly advertises ES256 and its PKCE requirements in the .well-known files. These are the defaults Kanidm ships with. Disabling them is possible but the commands are along the lines of "enable legacy crypto" etc., highly suggesting that users should only mess with this if necessary.

```bash 
wget -qO- https://kanidm:8443/oauth2/openid/splitpro/.well-known/openid-configuration 2>/dev/null 

"id_token_signing_alg_values_supported":["ES256"],
"code_challenge_methods_supported":["S256"]
```

The issue appears to lie in `openid-client` NextAuth v4 ships. 

`openid-client` defaults to RS256 and intentionally ignores `id_token_signing_alg_values_supported` from discovery. The expectation per spec is that the OP signs tokens with whatever alg the client was registered with. Since SplitPro passes credentials directly to openid-client with no additional configuration, and Kanidm defaults to ES256, the mismatch has to be resolved on the client side via `id_token_signed_response_alg`. The maintainer has expressed this as well: https://github.com/panva/openid-client/issues/115#issuecomment-418823617

As far as I can tell adding in the `checks` and `client` params is the intended way to deal with this, however, I have come across a different implementation as well: 
https://github.com/danny-avila/LibreChat/pull/5348 reads the discovered value and passes that on to the client. Happy to give this approach shot if it is more desirable.

docs/CONFIGURATION.md
mikimandoki: Fair enough - changes made.

src/server/auth.ts
krokosik: Yes, I like this approach, thanks for digging into this!

mikimandoki: Sorry for the delay on this. I did work on the above approach but it became a fairly significant change such that I don't feel confident shipping it anymore. I'll close this and submit a new PR with just the db crash fix, and keep the rest on a personal fork for now.

PR #690, opened 2026-06-22
feat: oauth autoredirect support

joshtgl:
# Description
Adds optional automatic redirect from the sign-in page to the first configured OAuth provider when `OAUTH_AUTO_REDIRECT=true`.

Let me know if any other changes/updates are needed.

# Demo
No UI screenshots included. Behavior change is:
- With `OAUTH_AUTO_REDIRECT=true` and at least one OAuth provider configured, visiting the sign-in page redirects immediately to the first configured OAuth provider.

# Checklist

- [X] I have read `CONTRIBUTING.md` in its entirety
- [X] I have performed a self-review of my own code
- [X] The last commit successfully passed pre-commit checks
- [X] Any AI code was thoroughly reviewed by me

## Summary by CodeRabbit

- **New Features**
  - Added an optional setting to automatically redirect sign-in when exactly one OAuth provider is configured.
  - Automatic redirection remains disabled by default and is skipped while verification or provider loading is active.
  - Sign-in failures during automatic redirection are reported with a notification.

- **Documentation**
  - Documented the automatic redirection setting and its configuration requirements.

- **Bug Fixes**
  - Added validation and warnings for unsupported automatic redirection configurations.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -67,6 +67,10 @@ Used for magic-link login and invites.
 
 ### OAuth providers
 
+#### Shared OAuth settings
+
+- `OAUTH_AUTO_REDIRECT`: Optional flag. When set to `true`, the sign-in page automatically redirects to the first configured OAuth provider. This applies to any configured OAuth provider, including Google, Authentik, Keycloak, and generic OIDC.
+
 #### Google
 
 - `GOOGLE_CLIENT_ID`
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -69,7 +69,7 @@ Used for magic-link login and invites.
 
 #### Shared OAuth settings
 
-- `OAUTH_AUTO_REDIRECT`: Optional flag. When set to `true`, the sign-in page automatically redirects to the first configured OAuth provider. This applies to any configured OAuth provider, including Google, Authentik, Keycloak, and generic OIDC.
+- `OAUTH_AUTO_REDIRECT`: Optional flag. When set to `true`, the sign-in page automatically redirects when exactly one OAuth provider is configured and no other authentication providers are enabled. This applies to Google, Authentik, Keycloak, and generic OIDC.
 
 #### Google
 
--- a/src/env.ts
+++ b/src/env.ts
@@ -71,6 +71,7 @@ export const env = createEnv({
     OIDC_CLIENT_SECRET: z.string().optional(),
     OIDC_WELL_KNOWN_URL: z.string().optional(),
     OIDC_ALLOW_DANGEROUS_EMAIL_LINKING: z.boolean().default(false),
+    OAUTH_AUTO_REDIRECT: z.boolean().optional(),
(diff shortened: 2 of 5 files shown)


src/env.ts
```diff
@@ -140,6 +141,7 @@ export const env = createEnv({
     OIDC_CLIENT_SECRET: process.env.OIDC_CLIENT_SECRET,
     OIDC_WELL_KNOWN_URL: process.env.OIDC_WELL_KNOWN_URL,
     OIDC_ALLOW_DANGEROUS_EMAIL_LINKING: Boolean(process.env.OIDC_ALLOW_DANGEROUS_EMAIL_LINKING),
+    OAUTH_AUTO_REDIRECT: 'true' === process.env.OAUTH_AUTO_REDIRECT,
```
krokosik: I would evaluate it in the same manner as the boolean above, namely wrap with `Boolean` to support values like `1` and keep consistency

src/pages/auth/signin.tsx
```diff
@@ -110,6 +111,15 @@ const Home: NextPage<{
     }
   }, [error, t]);
 
+  useEffect(() => {
+    if (oauthAutoRedirect && !showVerificationStep && providers.length > 0 && !isLoadingProviders) {
```
krokosik: I believe this should be only the case when there is exactly one provider specified and it's oauth. Otherwise there is no access to other options (especially email). Also let's display a console warning when the env var is set but the redirect does not happen due to multiple providers configured (not here, on startup)

docs/CONFIGURATION.md
```diff
@@ -67,6 +67,10 @@ Used for magic-link login and invites.
 
 ### OAuth providers
 
+#### Shared OAuth settings
+
+- `OAUTH_AUTO_REDIRECT`: Optional flag. When set to `true`, the sign-in page automatically redirects to the first configured OAuth provider. This applies to any configured OAuth provider, including Google, Authentik, Keycloak, and generic OIDC.
```
krokosik: Let's also add this env var to the .env.example (commented out)

krokosik (changes requested):

docs/CONFIGURATION.md
joshtgl: Added there and updated the description.

src/pages/auth/signin.tsx
joshtgl: Updated to require exactly 1 oauth provider and added a warning during the auth env validation.

src/env.ts
joshtgl: Wrapped like the other boolean env variables. Just a side note but it seems OIDC_ALLOW_DANGEROUS_EMAIL_LINKING should be wrapped the same way, it seems like any string would evaluate to true.

joshtgl:
I verified auto redirect on my local instance when the setting is enabled/disabled.

krokosik:
Great, thank you!

krokosik:
Ready to merge once the conflict is resolved @joshtgl

joshtgl:
It should be rebased and conflicts resolved.

PR #708, opened 2026-07-20
fix: prevent prisma crash on OIDC accounts with non-standard fields

mikimandoki:
# Description

Kanidm includes an `issued_token_type` field in its token response which isn't in the Prisma schema, causing a DB crash before the account could be created. Rather than adding another provider-specific case to `linkAccount`, this PR replaces the entire chain with a generic sanitisation pass using `Prisma.AccountScalarFieldEnum` as the source of truth. Any provider with non-standard fields now works without further changes.

# Checklist

See #682


krokosik:
Thank you and apologies for taking so long

PR #709, opened 2026-07-24
fix: harden cloud balance integrity

KMKoushik:
## Summary

- Make exact friendship settlements clear group projections through linked, reversible allocations instead of destructive zeroing.
- Keep partial friendship payments separate from group ledgers, matching Splitwise-style accounting contexts.
- Conserve integer cents deterministically across all editable split types and validate/normalize participant payloads at the server boundary.
- Reconstruct shares correctly when changing split type and hydrate edit state atomically, including category and attachment.
- Exclude settlements from group spending totals and show the dominant group-balance currency with a multi-currency marker.
- Add a lightweight Node test suite plus PostgreSQL integration coverage for balance and settlement lifecycles.

## Settlement projection

An exact non-group `SETTLEMENT` now snapshots every affected directional `GroupBalance` row into `SettlementAllocation` before clearing it. The allocation is linked to the settlement expense and contains the exact amount needed to reverse the projection.

- Editing an exact settlement to a partial payment restores the linked group amounts.
- Deleting a settlement restores allocations additively, preserving group activity recorded after the settlement.
- Financially unchanged edits retain their allocations instead of recalculating against later activity.
- Partial friendship payments and ordinary personal expenses do not alter group ledgers.
- Different currencies and unrelated friends remain isolated.
- If a group was deleted or a member left after settlement, its allocation is detached during reversal and the debt returns to the friendship ledger without recreating removed group state.
- Existing settlements are marked as legacy by the additive migration and cannot be edited or deleted because their destructively cleared group amounts cannot be reconstructed safely.

Balance-changing settlement operations, group expense creation/editing, group leave/delete, and friend deletion use serializable transactions with bounded retries. Group expense participants are revalidated inside the transaction so lifecycle checks cannot race allocation restoration or create inaccessible debt after a member leaves.

## Other integrity fixes

- Scope group projection changes to the correct pair and currency.
- Correct one-cent split drift such as USD 10 / 3 and normalize bounded drift from old clients.
- Reject malformed participant payloads, missing/duplicate payers, and invalid debt directions.
- Reject zero and non-finite SHARE totals without entering the cent-correction loop.
- Preserve edit category and attachment state.
- Exclude `SETTLEMENT` from group spending totals.
- Select the largest absolute nonzero group-summary currency and indicate multiple currencies.

## Verification

- `pnpm test`: 23 unit tests pass; 2 PostgreSQL tests skip unless `TEST_DATABASE_URL` is set.
- PostgreSQL 16 integration run: all 25 tests pass.
- Fresh database: all 8 cloud migrations apply successfully.
- `pnpm exec tsc --noEmit`: pass.
- `SKIP_ENV_VALIDATION=1 pnpm lint`: pass with existing branch warnings only.
- `SKIP_ENV_VALIDATION=1 pnpm build`: pass.
- Prisma validation, Prettier, and `git diff --check`: pass.

No historical balance repair is included. The migration is additive and preserves legacy settlement provenance conservatively.

## Summary by CodeRabbit

* **New Features**
  * Settlement expenses now preserve allocation details, allowing balances to be restored when settlements are edited or deleted.
  * Added improved support for personal and group expense processing, including safer balance updates.
  * Group balances now identify multi-currency groups with a visual indicator.
* **Bug Fixes**
  * Improved split calculations to prevent rounding discrepancies and reject invalid amounts.
  * Group and friend balance changes are now handled more reliably during leave, deletion, and expense updates.
* **Tests**
  * Added coverage for settlement restoration, balance summaries, and split accuracy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "postinstall": "prisma generate",
     "generate": "prisma generate",
     "lint": "next lint",
+    "test": "tsx --test src/tests/*.test.ts",
     "start": "sleep 3 && pnpm prisma:prod && next start",
     "start-with-latest-migrations": "prisma migrate deploy && next start",
     "d": "pnpm dx && pnpm dev",
--- a/prisma/migrations/20260726000000_add_settlement_allocations/migration.sql
+++ b/prisma/migrations/20260726000000_add_settlement_allocations/migration.sql
@@ -1,0 +1,22 @@
+-- AlterTable
+ALTER TABLE "Expense" ADD COLUMN "settlementAllocationVersion" INTEGER;
+
+-- CreateTable
+CREATE TABLE "SettlementAllocation" (
+    "settlementExpenseId" TEXT NOT NULL,
+    "groupId" INTEGER NOT NULL,
+    "currency" TEXT NOT NULL,
+    "userId" INTEGER NOT NULL,
+    "friendId" INTEGER NOT NULL,
+    "amount" INTEGER NOT NULL,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "SettlementAllocation_pkey" PRIMARY KEY ("settlementExpenseId", "groupId", "currency", "friendId", "userId")
+);
+
+-- CreateIndex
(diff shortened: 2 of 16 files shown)


KMKoushik:
@coderabbitai review

PR #717, opened 2026-08-02
fix: parse OIDC account linking flag as boolean

joshtgl:
# Description
Parse OIDC_ALLOW_DANGEROUS_EMAIL_LINKING as a JSON boolean so "false" does not enable dangerous account linking.

# Checklist

- [X] I have read `CONTRIBUTING.md` in its entirety
- [X] I have performed a self-review of my own code
- [X] The last commit successfully passed pre-commit checks
- [X] Any AI code was thoroughly reviewed by me

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of the dangerous email-linking configuration setting.
  * The setting now defaults to disabled when not explicitly configured.
  * Configuration values are interpreted consistently, with invalid values safely falling back to disabled.
  * This reduces the risk of unintended email account linking caused by ambiguous environment values.

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -70,7 +70,7 @@ export const env = createEnv({
     OIDC_CLIENT_ID: z.string().optional(),
     OIDC_CLIENT_SECRET: z.string().optional(),
     OIDC_WELL_KNOWN_URL: z.string().optional(),
-    OIDC_ALLOW_DANGEROUS_EMAIL_LINKING: z.boolean().optional(),
+    OIDC_ALLOW_DANGEROUS_EMAIL_LINKING: z.boolean().default(false),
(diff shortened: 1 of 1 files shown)


joshtgl:
Boolean handling is currently inconsistent and confusing with the JSON.parse wrapping and certain values can unexpectedly resolve to true. What would you think about a PR that updated all of the env boolean parsing logic to call a function such as:
```typescript
const parseEnvBoolean = (value: string | undefined): boolean => {
    return '1' === value || 'true' === value?.toLowerCase();
  };
```

krokosik:
Certainly! That was an oversight on my part, thanks for cleaning it up :)

PR #731, opened 2026-08-16
docs: fix broken star history chart

Dessalines39394:
The star history chart in the README currently fails to load because its previous source depends on the GitHub stargazer API, which now blocks those requests. This change points the chart to a working alternative so visitors can see the project's star history again. No other references are affected.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Star History chart image and link to use the current Star History website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,4 +176,4 @@ We are grateful for the support of our sponsors.
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


krokosik:
Nice, thank you

